PR #1, opened 2025-03-24
Add framework for cloudkit operator

larsks:
This pull request includes an operator generated using the [operator-sdk](https://sdk.operatorframework.io/), with resource fields in `api/v1alpha1/clusterorder_types.go` filled out to match (sort of) the fulfillment api. There are some impedance matches we're going to need to address around naming and around the list of conditions we want to track.

There's a lot of yack shaving in this pr to fix up linting configuration and linting errors in the operator-sdk generated code.


mhrivnak (approved):
We have some next changes we know we want to make, but I'm happy getting this merged now so we can address each of those details separately.

PR #8, opened 2025-03-26
Add cluster reference to ClusterOrder.status

larsks:
Add a `clusterReference` field to clusterOrder.status that records the namespace, cluster name, and service account name created by the controller in response to a ClusterOrder.

## Summary by CodeRabbit

- **New Features**
  - Enhanced order validations with a strict naming convention for template identifiers.
  - Added mandatory cluster reference details in the order status, requiring specific namespace and account information.
  - Refined order status phases with a clear set of allowed values to improve progress tracking.
- **Bug Fixes**
  - Removed deprecated condition tracking for node pool availability.
- **Documentation**
  - Updated Custom Resource Definition (CRD) to reflect new validation and structural requirements for `ClusterOrder`.


api/v1alpha1/clusterorder_types.go
```diff
@@ -63,20 +66,31 @@ const (
...
-	ClusterOrderConditionNodePoolAvailable ClusterOrderConditionType = "NodePoolAvailable"
-
 	// ClusterOrderConditionAvailable means the cluster is available
 	ClusterOrderConditionAvailable ClusterOrderConditionType = "Available"
 )
 
+// ClusterOrderClusterReferenceType contains a reference to the namespace created by this ClusterOrder
+type ClusterOrderClusterReferenceType struct {
+	// Namespace that contains the HostedCluster resource
+	ClusterNamespace   string `json:"clusterNamespace"`
```
mhrivnak, line 76: I don't think we need to repeat "cluster" here since it's already scoped by the parent field.

`instance.Status.ClusterReference.Namespace` is sufficient.

api/v1alpha1/clusterorder_types.go
```diff
@@ -63,20 +66,31 @@ const (
...
+	ClusterName        string `json:"clusterName"`
```
mhrivnak, line 77: How about `HostedClusterName` here since it's specific to that particular Kind? That would match the field below also.

api/v1alpha1/clusterorder_types.go
```diff
@@ -63,20 +66,31 @@ const (
...
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Type=string
+	// +kubebuilder:validation:Enum=Unknown;Accepted;Progressing;Failed;Ready
 	Phase ClusterOrderPhaseType `json:"phase,omitempty"`
 
 	// Conditions holds an array of metav1.Condition that describe the state of the ClusterOrder
 	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,1,rep,name=conditions"`
+
+	// Reference to the namespace that contains the HostedCluster resource
+	ClusterReference ClusterOrderClusterReferenceType `json:"clusterReference"`
```
mhrivnak, line 93: I think we want this whole field to be optional. It's a clean aggregation point; we either have a reference or we don't. It's common in k8s APIs to use a `nil` value to mean "we don't have any info on this sub-object".

IIRC making it optional just requires making this a pointer, and saying it's optional in the json metadata tag.

mhrivnak (changes requested):

api/v1alpha1/clusterorder_types.go
```diff
@@ -63,20 +66,31 @@ const (
...
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Type=string
+	// +kubebuilder:validation:Enum=Unknown;Accepted;Progressing;Failed;Ready
 	Phase ClusterOrderPhaseType `json:"phase,omitempty"`
 
 	// Conditions holds an array of metav1.Condition that describe the state of the ClusterOrder
 	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,1,rep,name=conditions"`
+
+	// Reference to the namespace that contains the HostedCluster resource
+	ClusterReference ClusterOrderClusterReferenceType `json:"clusterReference,omitempty"`
```
mhrivnak, line 93: I double-checked, and the k8s API guidelines do specify that optional fields should be a pointer. https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

mhrivnak (changes requested):

larsks:
@mhrivnak good catch; I've also marked `phase` and `conditions` as optional, which matches e.g. the Pod API.

mhrivnak (approved):

PR #10, opened 2025-03-27
Teach controller how to create and manage resources

larsks:
When a new ClusterOrder is created, update .status.clusterReference and
create dependent resources (namespace, service account, rolebinding). When
a ClusterOrder is deleted, delete the namespace (and transitively delete
all the resources contained in the namespace).

Because we're creating resources in a namespace different from the one in
which the ClusterOrder is created, we're relying on labels to manage watch
events and trigger reconciliation.

NB: This PR is on top of #8 and #9

## Summary by CodeRabbit

- **New Features**
  - Added a new field for role binding in ClusterOrder resources and updated the CRD schema to include it, ensuring enhanced configuration and validation.
  - Improved overall resource management, resulting in more robust orchestration of namespaces, service accounts, and role bindings.
  - Introduced a naming convention for CloudKit resources and a utility for generating unique namespace names.
  - Updated the operator image version to reflect the latest development branch.

- **Chores**
  - Refined configuration for code quality tools and dependency management.
  - Enhanced access permissions to better support resource operations.

### DIFF
--- a/internal/controller/clusterorder_controller.go
+++ b/internal/controller/clusterorder_controller.go
@@ -20,13 +20,44 @@ package controller
 import (
 	"context"
 
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/log"
+	controllerutil "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	cloudkitv1alpha1 "github.com/innabox/cloudkit-operator/api/v1alpha1"
 )
+
+// NewComponentFn is the type of a function that creates a required component
+type NewComponentFn func(context.Context, *cloudkitv1alpha1.ClusterOrder) (*appResource, error)
+
+type appResource struct {
+	object      client.Object
+	mutateFn    controllerutil.MutateFn
+	shouldExist bool
+}
+
+type component struct {
+	name string
+	fn   NewComponentFn
+}
+
+func (r *ClusterOrderReconciler) components() []component {
+	return []component{
+		{"Namespace", r.newNamespace},
+		{"ServiceAccount", r.newServiceAccount},
+		{"RoleBinding", r.newAdminRoleBinding},
+	}
+}
 
 // ClusterOrderReconciler reconciles a ClusterOrder object
 type ClusterOrderReconciler struct {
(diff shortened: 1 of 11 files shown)


internal/controller/clusterorder_controller.go
```diff
@@ -19,47 +19,261 @@ package controller
...
+	return ctrl.Result{}, nil
+}
+
+func (r *ClusterOrderReconciler) handleDelete(ctx context.Context, _ ctrl.Request, instance *cloudkitv1alpha1.ClusterOrder) (ctrl.Result, error) {
+	log := ctrllog.FromContext(ctx)
+	log.Info("Delete " + instance.GetName())
+
+	if controllerutil.ContainsFinalizer(instance, cloudkitFinalizer) {
+		// Can't do anything without status.clusterReference
+		if instance.Status.ClusterReference == nil {
```
mhrivnak, line 249: Along with my above comment, this should find the namespace ideally by label. Labels are part of a resource's identity, and it's the go-to way to associate resources when an owner reference isn't possible.

A guiding principle of k8s APIs is that you should be able to delete the status of a resource at any time, and the controller will reconstruct it from fresh observations.

Also, the status is just a point-in-time observation from the last time the controller was able to look. It makes no guarantees about the current state of the world.

mhrivnak (changes requested):

larsks:
@mhrivnak I've updated the code in the PR; let me know if this is in line with your comments.

internal/controller/clusterorder_controller.go
```diff
@@ -19,47 +19,257 @@ package controller
...
+		if err := r.Update(ctx, instance); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
+	// Do we have a clusterReference yet?
+	if instance.Status.ClusterReference == nil {
+		log.Info("Adding cluster reference")
+		nsName := namespaceNameFromInstance(instance)
+		instance.Status.ClusterReference = newClusterReference(nsName)
```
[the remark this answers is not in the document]
mhrivnak: I'm not sure what changed here, but I still don't see anything that is trying to observe the real existence of a HostedCluster prior to adding a reference to it in the status.

internal/controller/clusterorder_controller.go
```diff
@@ -19,47 +19,261 @@ package controller
...
+type NewComponentFn func(context.Context, *cloudkitv1alpha1.ClusterOrder) (*appResource, error)
+
+type appResource struct {
+	object      client.Object
+	mutateFn    controllerutil.MutateFn
+	shouldExist bool
+}
+
+type component struct {
+	name string
```
mhrivnak, line 53: nit: `kind` instead of `name` would be more consistent terminology with the k8s API conventions.

larsks: This really is meant to be a "name" and not a "kind"; it comes from [one of the examples you provided](https://github.com/saas-patterns/boutique-shop-operator/blob/6ad9b7112b15c14acdfc5dc0e8ac19aef066e7db/controllers/boutiqueshop_controller.go#L48). It's used as a human-readable descriptor in some log messages (and it makes the code a little easier to read to have the label there along with the function).

larsks:
> suspect that you forgot to actually push the updated code.

@mhrivnak Pushed to my development branch but not the PR :).

adriengentil (approved):

internal/controller/clusterorder_controller.go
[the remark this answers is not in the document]
larsks: Nothing in this PR monitors or responds to hosted clusters. We don't even set a watch on them until #11.

internal/controller/clusterorder_controller.go
mhrivnak: Yeah, the guy who made that example probably should have called it `kind` too. ;)

Anyone looking at controller logs, or especially controller code, should know what Group, Version and Kind are. That's the official terminology for defining a k8s API type.

https://book.kubebuilder.io/cronjob-tutorial/gvks.html
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#types-kinds

internal/controller/clusterorder_controller.go
```diff
@@ -20,46 +20,247 @@ package controller
...
+		).
+		Watches(
+			&rbacv1.RoleBinding{},
+			handler.EnqueueRequestsFromMapFunc(r.mapObjectToCluster),
+			builder.WithPredicates(labelPredicate(labelSelector)),
+		).
 		Complete(r)
 }
+
+func labelPredicate(label string) predicate.Predicate {
```
mhrivnak, line 120: You could use this instead: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/predicate#LabelSelectorPredicate

internal/controller/clusterorder_controller.go
mhrivnak: That said, it's not critical. And potentially when we get to creating multiple resources of the same Kind, for example multiple ConfigMaps or Secrets would be common, then we'll perhaps put something in this field that's a combination of kind and some other designation to give context.

internal/controller/clusterorder_controller.go
```diff
@@ -20,46 +20,247 @@ package controller
...
+		controllerutil.AddFinalizer(instance, cloudkitFinalizer)
+		if err := r.Update(ctx, instance); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
+	// Do we have a clusterReference yet?
+	if instance.Status.ClusterReference == nil {
+		log.Info("Adding cluster reference")
+		instance.Status.ClusterReference = newClusterReference()
```
mhrivnak, line 178: We should not add this until we have information to put inside it. That's the convention you'll see around the k8s API, and it's why we made sure this field is a pointer.

internal/controller/clusterorder_resources.go
```diff
@@ -0,0 +1,142 @@
...
+		Name:     "admin",
+	}
+
+	roleBinding := &rbacv1.RoleBinding{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      roleBindingName,
+			Namespace: namespaceName,
+			Labels:    commonLabelsFromOrder(instance),
+		},
+		Subjects: subjects,
```
mhrivnak, line 103: Typically anything outside the ObjectMeta should be added in the mutateFn, which always gets called regardless of the operation being a create or update. CreateOrUpdate will try to Get the object using its client, which if successful will overwrite contents of the object passed in, but IIRC optional or nil values won't get unset. So if you set desired state here, you can potentially end up with an object in the mutateFn that is a mix between the desired state and the actual current state. We want to ensure that it always represents just the current state when it pre-exists.

You are doing that now in the mutateFn, so you could just ditch these two lines since they're redundant anyway.

mhrivnak (commented):
I left a few more comments. Feel free to merge after resolving those, or we can talk more about them.

internal/controller/clusterorder_controller.go
larsks: That has actually already changed (https://github.com/larsks/cloudkit-operator/commit/d78276679aff6a7a4490e65343e2f75c0485d3ad#diff-c6507c1b735e1ac4b51a71ab5459385e88a41f5584de95efac80e0a90380429eL181).

larsks:
@mhrivnak this will merge once you approve; your requested change is currently blocking the merge.

internal/controller/clusterorder_controller.go
```diff
@@ -20,46 +20,236 @@ package controller
...
+			case controllerutil.OperationResultCreated:
+				log.Info("Created " + component.name)
+			case controllerutil.OperationResultUpdated:
+				log.Info("Updated " + component.name)
+			}
+
+			// apply any updates to status.clusterReference
+			if err := r.Status().Update(ctx, instance); err != nil {
+				return ctrl.Result{}, err
+			}
```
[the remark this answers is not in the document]
larsks: To be fixed in a future PR.

mhrivnak (approved):

PR #11, opened 2025-04-02
Call webhooks to create and tear down HostedCluster

larsks:
Adds support for calling a webhook to create the HostedCluster after the
controller has created the target namespace and associated resources, and
for a second webhook that is called to tear down the HostedCluster prior to
deleting the ClusterOrder.

This introduces two environment variables that can be used to set the
target URLs:

- `CLOUDKIT_CLUSTER_CREATE_WEBHOOK` -- the operator will post the JSON-serialized ClusterOrder to this URL after creating the target namespace, service account, and rolebinding.
- `CLOUDKIT_CLUSTER_DELETE_WEBHOOK` -- the operator will post the JSON-serialized ClusterOrder to this URL before deleting the target namespace.

**NB**: This PR is on top of #10 and will need to rebased after it merges.

## Summary by CodeRabbit

- **New Features**
  - Introduced webhook notifications for cluster creation and deletion via configurable environment variables.
  - Enhanced operator capabilities by automating the provisioning of namespaces, service accounts, and role bindings.
  - Expanded support for Hypershift API and updated the resource schema to include role binding details.
  - Added a new function for generating unique namespace names for `ClusterOrder` instances.
  - Implemented a new linter configuration to broaden static analysis checks.

- **Documentation**
  - Updated the README with a clearer operator overview and detailed configuration instructions.

- **Chores**
  - Adjusted the controller image reference and tag for development consistency.
  - Updated Go version and dependencies to ensure compatibility with the latest features and fixes.


internal/controller/clusterorder_webhook.go
```diff
@@ -0,0 +1,45 @@
...
+	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
+
+	cloudkitv1alpha1 "github.com/innabox/cloudkit-operator/api/v1alpha1"
+)
+
+func triggerWebHook(ctx context.Context, url string, instance *cloudkitv1alpha1.ClusterOrder) error {
+	log := ctrllog.FromContext(ctx)
+	log.Info("Triggering webhook " + url)
+
+	jsonData, err := json.Marshal(instance)
```
mhrivnak, line 20: Fine for now perhaps, but I think we'll want to give this a minimal envelope so we can pass other stuff in addition to the cluster order. For example, maybe we'd pass a token or other resources with useful context. Up to you if you want to establish that now or cross that bridge when it becomes useful.

larsks: I think I'll wait until we have a concrete requirement. @adriengentil had some thoughts about tokens and impersonation that may mean we don't need to worry about passing tokens around.

adriengentil: As you said, I don't think we need it immediately. Another advantage of the envelope would be to decouple ourselves from HTTP webhooks as we rely on the URL path to route the action (create/delete), it would aloow us to be able to use other communication ways (kafka for example).

internal/controller/clusterorder_webhook.go
```diff
@@ -0,0 +1,45 @@
...
+
+	req, err := http.NewRequest("POST", url, bytes.NewBuffer(jsonData))
+	if err != nil {
+		return fmt.Errorf("failed to create request: %w", err)
+	}
+
+	req.Header.Set("Content-Type", "application/json")
+
+	client := &http.Client{Timeout: 10 * time.Second}
+	resp, err := client.Do(req)
```
adriengentil, line 33: no need to retry if the call to the webhook fails because we will try to reconcile again later, is my understanding correct?

adriengentil (approved):
I have one question, otherwise it looks good to me!

internal/controller/clusterorder_webhook.go
larsks: I believe that returning a non-`nil` error from the Reconcile function triggers another run of the reconcile loop.

PR #15, opened 2025-04-03
Don't re-implement maps.Copy()

larsks:


tzumainn (approved):
yep, I guess that's what maps.Copy does!

PR #16, opened 2025-04-04
Remove unused code and update log messages

larsks:
This is a maintenance change that doesn't have any operational impact on the code. It is primarily code cleanup, log message improvements, and de-emphasizing docker in the Makefile.

### DIFF
--- a/internal/controller/clusterorder_controller.go
+++ b/internal/controller/clusterorder_controller.go
@@ -19,11 +19,11 @@ package controller
 
 import (
 	"context"
+	"fmt"
 
 	hypershiftv1beta1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


tzumainn (approved):

adriengentil (approved):

internal/controller/clusterorder_controller.go
```diff
@@ -81,19 +76,25 @@ func newClusterReference() *v1alpha1.ClusterOrderClusterReferenceType {
...
-	_ = ctrllog.FromContext(ctx)
+	log := ctrllog.FromContext(ctx)
 
 	instance := &v1alpha1.ClusterOrder{}
 	err := r.Client.Get(ctx, req.NamespacedName, instance)
 	if err != nil {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
+	log.Info(fmt.Sprintf("Start reconcile for %s", instance.GetName()))
```
adriengentil, line 87: there's no `log.Infof` to avoid the nested `fmt.Sprintf`?

larsks: Not in [this log package](https://pkg.go.dev/github.com/go-logr/logr#Logger).

adriengentil: https://github.com/go-logr/logr?tab=readme-ov-file#but-i-really-want-to-use-a-format-string

looks like this logger wants to push another way to log things

PR #26, opened 2025-04-16
Add nodeRequests field to CRD

larsks:
This commit adds the spec.nodeRequests attribute to the ClusterOrder CRD.

    $ kubectl explain clusterorder.spec.nodeRequests

    FIELD: nodeRequests <[]Object>

    DESCRIPTION:
        NodeRequests defines the types of nodes and number of each type of
        node that will be used to build the cluster. This value is optional
        and if not provided will be filled in with template-provided
        defaults. The selected template may limit what node types you can
        request.

    FIELDS:
      numberOfNodes <integer> -required-
        NumberOfNodes describes the number of nodes you want of the given resource
        class

      resourceClass <string> -required-
        ResourceClass describes the type of node you are requesting

Addresses innabox/issues#100
Related to https://github.com/innabox/cloudkit-aap/pull/24, which includes template metadata for default `nodeRequests` values


jhernand (approved):
Looks good to me.

PR #27, opened 2025-04-24
Send updates to the fulfillment service

jhernand:
This patch adds a new _Feedback_ controller that watches the Kubernetes `ClusterOrder` object and sends sends updates to the fulfillment service.


jhernand:
There are many files in this pull request, but most of them are automatically generated because I needed to import the gRPC API. Focus your review in the `internal/controller/feedback_controller.go` file, as that is were all the logic is.

larsks (approved):
This looks fine. The feedback controller only monitors the clusterorder...should we adopt a similar model (i.e., using a dedicated controller) for watching the HostedCluster and updating the ClusterOrder status?

jhernand:
> This looks fine. The feedback controller only monitors the clusterorder...should we adopt a similar model (i.e., using a dedicated controller) for watching the HostedCluster and updating the ClusterOrder status?

This controller is already fetching the HostedCluster (to get the API and console URLs) so I'd say it would be better to modify it so that it also whatches HostedClusters, no need for I new controller I think.

It would be even better if all the required information (API URL, console URL, location of the admin Kubeconfig secret, status of node sets, etc) was in the status of the Kubernetes ClusterOrder object. That way this controller wouldn't need to be aware of the existence of the "HostedCluster" object. But my understanding is that we have already decided to not do that.

PR #31, opened 2025-05-07
Add nodeCounts field to status

larsks:
This adds a nodeCounts field to the ClusterOrder status:

    status:
      nodeCounts:
      - resourceClass: foo
        numberOfNodes: 2
      - resourceClass: bar
        numberOfNodes: 1

Closes innabox/issues#117


jhernand:
Looks good, thanks @larsks.

mhrivnak (changes requested):
Copying from my comment on the issue: I'd suggest something like this, which leaves the door open for additional status fields per resource class beyond just the raw node count.

```
status:
  resourceClasses:
  - name: foo
    nodeCount: 4
```

Whether you look at NodePools, MachineConfigPools, MachineSets, they all tend to have additional status fields that can be useful beyond just the count, such as sub-dividing into how many are "ready". I don't think we need to add additional fields now, but setting up the data structure to accommodate more fields seems wise and would be idiomatic for a k8s API.

WDYT?

larsks:
> Copying from my comment on the issue: I'd suggest something like this, which leaves the door open for additional status fields per resource class beyond just the raw node count.

@mhrivnak I think that's pretty much exactly what this PR does? Instead of `resourceClasses` we've called it `nodeCounts` (to parallel `nodeRequests` in `spec`), but otherwise the structure is the same:

```
status:
  nodeCounts:
  - resourceClass: fc430
    numberOfNodes: 2
```

You could easily add additional attributes to this if that becomes necessary.

larsks:
> The difference is that the array should not necessarily just be limited to "node counts", but should be status for each resource class ... which includes node counts and likely other additional status.

Can you suggest what additional information you might expect to see here? I would expect information about a particular agent to be associated with the agent object; I'm not sure that we would be tracking any attributes at the resource class level. 

I'm hoping to keep this simple and have it reflect the status of `spec.nodeRequests`.

mhrivnak:
I think these are equally simple, but one is future-proof.

For examples of what sort of information we might add in the future, the NodePool resource captures a lot of information in its own status, some of which could be useful to propagate in one form or another: https://github.com/openshift/hypershift/blob/9fd076913fdaadba0741fa70b3feb033e1c3c0cb/support/conditions/conditions.go#L96-L110

This could be a place to express the status of quota usage.

Or like a ReplicaSet shows additional counters for how many pods are in a particular state, in addition to the total count, we may want to show here how many nodes are actually part of the cluster, vs in the process of joining or leaving, etc.

Those are all just ideas and examples. All I'm suggesting for now is that we use a data structure that leaves the door open for our future selves to add information.

Looking at the [API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps) for lists vs. maps, it's worth noting the guidance "the convention is to use a list of subobjects containing name fields". It's not just making an argument against maps, but also telling us how to make lists. And that's the pattern you'll find over and over in the k8s API where there's a list of objects.

To your point about wanting to reflect the status of `spec.nodeRequests`, why not as a starting point just use the exact same data structure in the spec and status? It doesn't have a name field, but we could think over that separately. 

```
status:
  nodeRequests:
  - resourceClass: foo
    numberOfNodes: 4
```

larsks:
> To your point about wanting to reflect the status of spec.nodeRequests, why not as a starting point just use the exact same data structure in the spec and status?

We are...explicitly:

```
	// NodeCount shows how many nodes are currently associated with the ClusterOrder
	NodeCounts []NodeRequest `json:"nodeCounts,omitempty"
```

It just has a different name because it's no longer a request.

mhrivnak:
Ok I see. I would consider `status.nodeRequests` to be a fine way to tell me the status of my requests. I don't think there's a need to rename it.

If anyone really wants to rename it, I'll just argue to keep "counts" out of the name, so that we can add more info to the struct later if we want.

I might suggest `status.nodes` if `Node` wasn't already a well-known API.

larsks:
@mhrivnak if you're happy with this, you'll need to retract your "requested change" or approve it :)

mhrivnak (approved):

PR #32, opened 2025-05-08
Copy actual replicas from node pool to cluster order status

jhernand:
This patch changes the controller so that it copies the `status.replicas` field of the `NodePool` object into the `status.nodeCounts[].numberOfNodes` field of the `ClusterOrder` object.

Note that currently there is no way to know what is the `resourceClass` that corresponds to a node pool, so the best we can do is assume that there will be only one item in the `status.nodeRequests` field of the `CluserOrder` object, and only one corresponding node pool.

Note that this is on top of #31.

### DIFF
--- a/internal/controller/clusterorder_controller.go
+++ b/internal/controller/clusterorder_controller.go
@@ -173,6 +173,11 @@ func (r *ClusterOrderReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			handler.EnqueueRequestsFromMapFunc(r.mapObjectToCluster),
 			builder.WithPredicates(labelPredicate),
 		).
+		Watches(
+			&hypershiftv1beta1.NodePool{},
+			handler.EnqueueRequestsFromMapFunc(r.mapObjectToCluster),
+			builder.WithPredicates(labelPredicate),
+		).
 		Complete(r)
 }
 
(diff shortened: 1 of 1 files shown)


larsks:
> Note that currently there is no way to know what is the resourceClass that corresponds to a node pool

I've opened https://github.com/innabox/issues/issues/119 to track this.

internal/controller/clusterorder_controller.go
```diff
@@ -234,24 +239,108 @@ func (r *ClusterOrderReconciler) handleUpdate(ctx context.Context, _ ctrl.Reques
...
+	instance.SetClusterReferenceHostedClusterName(name)
+	instance.SetStatusCondition(v1alpha1.ConditionControlPlaneCreated, metav1.ConditionTrue, "", v1alpha1.ReasonAsExpected)
+
+	if controlPlaneIsAvailable(hc) {
+		instance.SetStatusCondition(v1alpha1.ConditionControlPlaneAvailable, metav1.ConditionTrue, "", v1alpha1.ReasonAsExpected)
+	}
+
+	// Fetch the node pools and handle them:
+	nodePools := &hypershiftv1beta1.NodePoolList{}
+	err := r.Client.List(ctx, nodePools, client.MatchingLabels{
```
larsks, line 260: This may be a dumb question, or maybe I shouldn't be looking at code this late at night but: how does `client.List` know in which namespace to look for resources? Or is this performing a lookup across all namespaces? It obviously works (we're already doing this elsewhere in the code), but looking at it just now it seems odd.

The [documentation](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client#Reader) says only:

```
	// List retrieves list of objects for a given namespace and list options. On a
	// successful call, Items field in the list will be populated with the
	// result returned from the server.
```

But what is "the given namespace"? I don't think we're using a namespaced client, and even if we were it would be bound to the wrong namespace.

I *think* we're listing resources across all namespaces to which we have access. Fortunately, the label selector will prevent us from any erroneous matches, but should we be restricting all the List methods to a single namespace?

larsks (approved):
I've rebased the PR; it otherwise seems fine. I left a question because there's something I'm missing, but it's not particular to this PR.

internal/controller/clusterorder_controller.go
jhernand: This is indeed listing across all namespaces. To list in a particular namespace you need to add the `InNamespace(...)` option, something like this:

```go
err := r.Client.List(ctx, nodePools, client.InNamespace(...))
```

I will add that.

jhernand: Done.

PR #35, opened 2025-05-12
Enable the feedback controller

jhernand:
This patch adds the `--fulfillment-server-address` flag to the command line of the controller manager, so that the fulfillment controller will be enabled.

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -63,6 +63,9 @@ spec:
         args:
           - --leader-elect
           - --health-probe-bind-address=:8081
+          - --grpc-insecure
+          - --grpc-token-file=/var/run/secrets/kubernetes.io/serviceaccount/token
+          - --fulfillment-server-address=fulfillment-api.innabox.svc:8000
         image: controller:latest
         imagePullPolicy: Always
         name: manager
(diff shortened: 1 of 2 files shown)


config/manager/manager.yaml
```diff
@@ -63,6 +63,9 @@ spec:
         args:
           - --leader-elect
           - --health-probe-bind-address=:8081
+          - --grpc-insecure
+          - --grpc-token-file=/var/run/secrets/kubernetes.io/serviceaccount/token
```
tzumainn, line 67: Should this (and the value below) be part of a configuration file or something? It does feel a little strange to have this hardcoded for our specific testing environment.

larsks: Yeah, this should be loaded from environment variables like we're doing with other configuration options (e.g., `CLOUDKIT_CLUSTER_CREATE_WEBHOOK`). I'll update the PR.

larsks (changes requested):

jhernand:
> @jhernand I've pushed some changes to this PR so that we get default values for the command line options from the environment. Does that seem reasonable?

Yes, it seems reasonable to me.

> Also, do we need to be using `--grpc-insecure`? If we're connecting to another service on the same cluster, we should already trust that certificate.

The certificate that we use currently is signed by the service signer, and that isn't automatically trusted by the Go HTTP client. We can change the client so that it trusts it by default, or add flags (or environment variables) to make it configurable, or else use this insecure flag.

tzumainn (approved):

larsks (approved):
Looks good to me. We can address certificate trust in a future pr.

PR #36, opened 2025-05-12
Copy all source files into container build environment

larsks:
The Containerfile was previously too granular and was not copying the
complete source tree into the build environment, resulting in failed
builds.


tzumainn (approved):
Looks good!

Issue #38, opened 2025-05-12
Avoid triggering webhooks too frequently


larsks:
Add logic to the [cloudkit kubernetes controller](https://github.com/larsks/cloudkit-operator/) to introduce a minimum interval between subsequent triggers of a given webhook. This should be a fixed interval, configurable at runtime.

PR #40, opened 2025-05-13
Resolve issue #38 - avoid too fast webhooks

jkary:
Added MinimumRequestInterval to ClusterOrder with a default of 5m.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"time"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
@@ -72,6 +73,7 @@ func main() {
 	var grpcInsecure bool
 	var grpcTokenFile string
 	var fulfillmentServerAddress string
+	var minimumRequestInterval string
 	var tlsOpts []func(*tls.Config)
 	flag.StringVar(&metricsAddr, "metrics-bind-address", "0", "The address the metrics endpoint binds to. "+
 		"Use :8443 for HTTPS or :8080 for HTTP, or leave as 0 to disable the metrics service.")
@@ -106,6 +108,12 @@ func main() {
 		"fulfillment-server-address",
 		os.Getenv("CLOUDKIT_FULFILLMENT_SERVER_ADDRESS"),
 		"Address of the fulfillment server.",
+	)
+	flag.StringVar(
+		&minimumRequestInterval,
+		"minimum-request-interval",
+		os.Getenv("CLOUDKIT_MINIMUM_REQUEST_INTERVAL"),
+		"Minimum amount of time between calls to the same webook url",
 	)
 	opts := zap.Options{
 		Development: true,
@@ -208,12 +216,24 @@ func main() {
 		setupLog.Info("gRPC connection to fulfillment service is disabled")
 	}
 
+	// No minimumRequestInterval means no rate limiting
+	if minimumRequestInterval == "" {
+		minimumRequestInterval = "0"
+	}
+
+	interval, err := time.ParseDuration(minimumRequestInterval)
+	if err != nil {
+		setupLog.Error(err, "Invalid minimum request interval.")
+		os.Exit(1)
+	}
+
 	if err = (controller.NewClusterOrderReconciler(
 		mgr.GetClient(),
 		mgr.GetScheme(),
 		os.Getenv("CLOUDKIT_CLUSTER_CREATE_WEBHOOK"),
 		os.Getenv("CLOUDKIT_CLUSTER_DELETE_WEBHOOK"),
 		os.Getenv("CLOUDKIT_CLUSTER_ORDER_NAMESPACE"),
+		interval,
 	)).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "ClusterOrder")
 		os.Exit(1)
--- a/internal/controller/clusterorder_controller.go
+++ b/internal/controller/clusterorder_controller.go
@@ -20,6 +20,7 @@ package controller
 import (
 	"context"
 	"fmt"
+	"time"
 
 	hypershiftv1beta1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
 	corev1 "k8s.io/api/core/v1"
@@ -64,10 +65,11 @@ func (r *ClusterOrderReconciler) components() []component {
 // ClusterOrderReconciler reconciles a ClusterOrder object
 type ClusterOrderReconciler struct {
 	client.Client
-	Scheme                *runtime.Scheme
-	CreateClusterWebhook  string
-	DeleteClusterWebhook  string
-	ClusterOrderNamespace string
+	Scheme                 *runtime.Scheme
+	CreateClusterWebhook   string
+	DeleteClusterWebhook   string
+	ClusterOrderNamespace  string
+	MinimumRequestInterval time.Duration
 }
 
 func NewClusterOrderReconciler(
@@ -76,16 +78,20 @@ func NewClusterOrderReconciler(
 	createClusterWebhook string,
 	deleteClusterWebhook string,
 	clusterOrderNamespace string,
+	minimumRequestInterval time.Duration,
 ) *ClusterOrderReconciler {
+
 	if clusterOrderNamespace == "" {
 		clusterOrderNamespace = defaultClusterOrderNamespace
 	}
+
 	return &ClusterOrderReconciler{
-		Client:                client,
-		Scheme:                scheme,
-		CreateClusterWebhook:  createClusterWebhook,
-		DeleteClusterWebhook:  deleteClusterWebhook,
-		ClusterOrderNamespace: clusterOrderNamespace,
+		Client:                 client,
+		Scheme:                 scheme,
+		CreateClusterWebhook:   createClusterWebhook,
+		DeleteClusterWebhook:   deleteClusterWebhook,
+		ClusterOrderNamespace:  clusterOrderNamespace,
+		MinimumRequestInterval: minimumRequestInterval,
 	}
 }
 
@@ -256,9 +262,18 @@ func (r *ClusterOrderReconciler) handleUpdate(ctx context.Context, _ ctrl.Reques
 		val, exists := instance.Annotations[cloudkitManagementStateAnnotation]
 		if exists && val == "manual" {
 			log.Info(fmt.Sprintf("Not triggering create webhook for ClusterOrder %s because of annotation", instance.GetName()))
-		} else if err := triggerWebHook(ctx, url, instance); err != nil {
-			log.Error(err, fmt.Sprintf("Failed to trigger webhook %s: %v", url, err))
-			return ctrl.Result{Requeue: true}, nil
+		} else {
+			remainingTime, err := triggerWebHook(ctx, url, instance, r.MinimumRequestInterval)
+			if err != nil {
+				log.Error(err, fmt.Sprintf("Failed to trigger webhook %s: %v", url, err))
+				return ctrl.Result{Requeue: true}, nil
+			}
+
+			// Verify if we are within the minimum request window
+			if remainingTime != 0 {
+				log.Info(fmt.Sprintf("Request %s is within minimumRequestInterval", url))
+				return ctrl.Result{RequeueAfter: remainingTime}, nil
+			}
 		}
 	}
 
@@ -436,9 +451,16 @@ func (r *ClusterOrderReconciler) handleDelete(ctx context.Context, _ ctrl.Reques
 				val, exists := instance.Annotations[cloudkitManagementStateAnnotation]
 				if exists && val == "manual" {
 					log.Info(fmt.Sprintf("Not triggering delete webhook for ClusterOrder %s because of annotation", instance.GetName()))
-				} else if err := triggerWebHook(ctx, url, instance); err != nil {
-					log.Error(err, fmt.Sprintf("Failed to trigger webhook %s: %v", url, err))
-					return ctrl.Result{Requeue: true}, nil
+				} else {
+					remainingTime, err := triggerWebHook(ctx, url, instance, r.MinimumRequestInterval)
+					if err != nil {
+						log.Error(err, fmt.Sprintf("Failed to trigger webhook %s: %v", url, err))
+						return ctrl.Result{Requeue: true}, nil
+					}
+
+					if remainingTime != 0 {
+						return ctrl.Result{RequeueAfter: remainingTime}, nil
+					}
 				}
 			}
 			return ctrl.Result{}, err
--- a/internal/controller/clusterorder_webhook.go
+++ b/internal/controller/clusterorder_webhook.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"sync"
 	"time"
 
 	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
@@ -13,18 +14,83 @@ import (
 	cloudkitv1alpha1 "github.com/innabox/cloudkit-operator/api/v1alpha1"
 )
 
-func triggerWebHook(ctx context.Context, url string, instance *cloudkitv1alpha1.ClusterOrder) error {
+type InflightRequest struct {
+	createTime time.Time
+}
+
+var (
+	inflightRequestsLock *sync.RWMutex              = &sync.RWMutex{}
+	inflightRequests     map[string]InflightRequest = make(map[string]InflightRequest)
+)
+
+func checkForExistingRequest(ctx context.Context, url string, minimumRequestInterval time.Duration) time.Duration {
+	var delta time.Duration
+
+	inflightRequestsLock.RLock()
+	request, ok := inflightRequests[url]
+	if ok {
+		delta = time.Since(request.createTime)
+		if delta >= minimumRequestInterval {
+			delta = 0
+		}
+	}
+	inflightRequestsLock.RUnlock()
+	purgeExpiredRequests(ctx, minimumRequestInterval)
+	return delta
+}
+
+func addInflightRequest(ctx context.Context, url string, minimumRequestInterval time.Duration) {
 	log := ctrllog.FromContext(ctx)
+	inflightRequestsLock.Lock()
+	inflightRequests[url] = InflightRequest{
+		createTime: time.Now(),
+	}
+	log.Info(fmt.Sprintf("Adding %s to cache", url))
+	inflightRequestsLock.Unlock()
+	purgeExpiredRequests(ctx, minimumRequestInterval)
+}
+
+func purgeExpiredRequests(ctx context.Context, minimumRequestInterval time.Duration) {
+	var expiredRequests = []string{}
+
+	log := ctrllog.FromContext(ctx)
+	inflightRequestsLock.RLock()
+	for url, request := range inflightRequests {
+		if delta := time.Since(request.createTime); delta > minimumRequestInterval {
+			log.Info(fmt.Sprintf("Found %s with a delta of %d", url, delta))
+			expiredRequests = append(expiredRequests, url)
+		}
+	}
+	inflightRequestsLock.RUnlock()
+
+	if len(expiredRequests) > 0 {
+		inflightRequestsLock.Lock()
+		for _, url := range expiredRequests {
+			log.Info(fmt.Sprintf("Deleting %s from cache", url))
+			delete(inflightRequests, url)
+		}
+		inflightRequestsLock.Unlock()
+	}
+}
+
+func triggerWebHook(ctx context.Context, url string, instance *cloudkitv1alpha1.ClusterOrder, minimumRequestInterval time.Duration) (time.Duration, error) {
+
+	log := ctrllog.FromContext(ctx)
+
+	if delta := checkForExistingRequest(ctx, url, minimumRequestInterval); delta != 0 {
+		return delta, nil
+	}
+
 	log.Info("Triggering webhook " + url)
 
 	jsonData, err := json.Marshal(instance)
 	if err != nil {
-		return fmt.Errorf("failed to marshal JSON: %w", err)
+		return 0, fmt.Errorf("failed to marshal JSON: %w", err)
 	}
 
 	req, err := http.NewRequest("POST", url, bytes.NewBuffer(jsonData))
 	if err != nil {
-		return fmt.Errorf("failed to create request: %w", err)
+		return 0, fmt.Errorf("failed to create request: %w", err)
 	}
 
 	req.Header.Set("Content-Type", "application/json")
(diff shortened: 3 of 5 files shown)


api/v1alpha1/clusterorder_types.go
```diff
@@ -123,8 +123,9 @@ type ClusterOrder struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec   ClusterOrderSpec   `json:"spec,omitempty"`
-	Status ClusterOrderStatus `json:"status,omitempty"`
+	Spec                   ClusterOrderSpec   `json:"spec,omitempty"`
+	Status                 ClusterOrderStatus `json:"status,omitempty"`
+	MinimumRequestInterval string             `json:"minduration,omitempty"`
```
mhrivnak, line 128: This struct is the API. We want to specify the interval as a configuration option for the process itself, which is running the controller that happens to implement the API. Instead of trying to use the API for this, we should probably receive this config option similar to some other ones seen in main.go: https://github.com/innabox/cloudkit-operator/blob/50bce09/cmd/main.go#L104-L109

internal/controller/clusterorder_webhook.go
```diff
@@ -13,8 +13,34 @@ import (
 	cloudkitv1alpha1 "github.com/innabox/cloudkit-operator/api/v1alpha1"
 )
 
+var instanceTracking map[string]time.Time
```
mhrivnak, line 16: I'd consider a more meaningful name here, like `lastCallByURL`

internal/controller/clusterorder_webhook.go
```diff
@@ -13,8 +13,34 @@ import (
...
+		}
+
+		if instance.MinimumRequestInterval != "" {
+			log.Info("Invalid minimum request interval.  Setting it to the default.", "MinimumRequestInterval", DefaultRequestInterval, "")
+			duration, _ = time.ParseDuration(DefaultRequestInterval)
+		}
+
+		minInterval := int64(duration.Seconds())
+		delta := int64(time.Since(lastUpdate).Seconds())
+		if delta < minInterval {
```
mhrivnak, line 36: Why do these need to be `int64`? Can't we compare them natively?

internal/controller/clusterorder_webhook.go
```diff
@@ -13,8 +13,34 @@ import (
...
+			log.Info("Tring to call webhook too soon.", "url", url, "minInterval", minInterval, "delta", int(delta), "lastUpdate", lastUpdate.String())
+			return fmt.Errorf("request inside the minimum request interval of %s", instance.MinimumRequestInterval)
```
mhrivnak, line 38: I think it would a good idea to also return the delta here, so that the `Reconcile` function can return with a `RequeueAfter` value, thus ensuring that the reconciler has another chance to do whatever it was trying to do.

internal/controller/clusterorder_webhook.go
```diff
@@ -13,8 +13,34 @@ import (
...
+		}
+	} else {
+		instanceTracking[url] = time.Now()
```
mhrivnak, line 41: We shouldn't set this until we're sure that the request was successful. If the webhook is down for example and we get some TCP or HTTP error when calling it, we wouldn't want to wait 5 minutes to re-try.

mhrivnak (changes requested):

internal/controller/clusterorder_webhook.go
larsks: @mhrivnak Does Kuberntes coalesce multiple `RequeAfter` requests?

api/v1alpha1/clusterorder_types.go
larsks: We had discussed perhaps making this a field of the `ClusterOrderReconciler`, which is defined in `internal/controller/clusterorder_controller.go`. That's currently where other runtime options are stored (like the webhook urls).

internal/controller/clusterorder_webhook.go
mhrivnak: Good question. In this case it would be controller-runtime itself doing that, and it does so using a priority queue that says it "internally de-duplicates all items": https://github.com/kubernetes-sigs/controller-runtime/blob/3c3fd3e/pkg/controller/priorityqueue/priorityqueue.go#L31

api/v1alpha1/clusterorder_types.go
jkary: This makes sense.  It seemed to make sense to fit in the ClusterOrder to avoid adding a new function parameter, but I agree with the separation of concerns using ClusterOrderReconciler.

internal/controller/clusterorder_webhook.go
jkary: I changed it to integers.   time.Duration.Seconds() returns a float64 and I was concerned Go would force an int64.

internal/controller/clusterorder_webhook.go
jkary: I am not familiar with the controller runtime repo.  I agree we should return the delta time needed before allowing another request.  Are you thinking we should use a custom error type or just return a tuple?

internal/controller/clusterorder_webhook.go
mhrivnak: How about `if time.Since(lastupdate) < duration` ?

internal/controller/clusterorder_controller.go
```diff
@@ -311,7 +315,7 @@ func (r *ClusterOrderReconciler) handleDelete(ctx context.Context, _ ctrl.Reques
 	if hc, err := r.findHostedCluster(ctx, instance); hc != nil {
 		log.Info(fmt.Sprintf("Waiting for HostedCluster %s to delete", hc.GetName()))
 		if url := r.DeleteClusterWebhook; url != "" {
-			if err := triggerWebHook(ctx, url, instance); err != nil {
+			if err := triggerWebHook(ctx, url, instance, r.MinimumRequestInterval); err != nil {
```
mhrivnak, line 318: The fact that the webhook logic wants to wait isn't really an error. In fact we probably expect that to be the case a lot of the time as a normal workflow. That's supported by the logic below, which logs an error and returns with an immediate requeue if it sees an error returned. That's fairly normal behavior, but not what we want in case we just want a delay.

If it fails to call the webhook, we should just requeue so it'll try again, as we see already implemented in the two lines below. controller-runtime will add automatic backoff.

If it wants to just delay calling the webhook, then we want this function (`handleCreate()` or `handleUpdate()`) to instead return `ctrl.Result{RequeueAfter: someDuration}`.

The clearest way to distinguish these two scenarios is to have `triggerWebHook` return two values.

```
// triggerWebHook calls an external webhook that implements additional capabilities.
// it returns an error only when calling the webhook failed or was not possible for some
// unexpected reason.
// it returns a duration when the specific URL has already been called recently; the caller
// should wait at least that much time and then call the function again with the same URL.
func triggerWebHook(ctx ...) (time.Duration, error) {
```

internal/controller/clusterorder_webhook.go
```diff
@@ -13,8 +13,41 @@ import (
...
+}
+
+func (e *ErrMinInterval) Error() string {
+	return fmt.Sprintf("request inside the minimum request interval of %v", string(e.RemainingTime))
+}
+
+func triggerWebHook(ctx context.Context, url string, instance *cloudkitv1alpha1.ClusterOrder, minimumRequestInterval string) error {
 	log := ctrllog.FromContext(ctx)
+
+	if inflightTime, found := inflightRequests[url]; found {
```
mhrivnak, line 31: We'll also want to use a lock to read or update this, since multiple goroutines could be running Reconcile concurrently. Or use a concurrency-safe cache with TTL like Adrien suggested.

mhrivnak (changes requested):

internal/controller/clusterorder_webhook.go
```diff
@@ -13,8 +13,41 @@ import (
...
+
+type ErrMinInterval struct {
+	RemainingTime string
+}
+
+func (e *ErrMinInterval) Error() string {
+	return fmt.Sprintf("request inside the minimum request interval of %v", string(e.RemainingTime))
+}
+
+func triggerWebHook(ctx context.Context, url string, instance *cloudkitv1alpha1.ClusterOrder, minimumRequestInterval string) error {
```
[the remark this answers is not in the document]
jkary: I see that I have some reading to do about kubernetes controllers.

internal/controller/clusterorder_webhook.go
jkary: Refactoring inflightRequests to use cacheing which has locking mechanism builtin.

internal/controller/clusterorder_controller.go
jkary: Makes sense now.  I have made the corrections.

internal/controller/clusterorder_webhook.go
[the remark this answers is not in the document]
jkary: Agreed.

internal/controller/clusterorder_webhook.go
```diff
@@ -5,41 +5,55 @@ import (
...
+
 	log := ctrllog.FromContext(ctx)
+	minDelta, _ := time.ParseDuration(minimumRequestInterval)
+
+	if item, err := (*inflightRequests).Value(url); err != nil {
+		inflightTime := item.Data().(time.Time)
+		delta := time.Since(inflightTime)
+		if delta < minDelta {
+			return delta, fmt.Errorf("Tring to call webhook too soon. url: %s minInterval: %s", url, minimumRequestInterval)
+		}
```
adriengentil, line 28: is this check required? I would expect the cache to evict the entry if the TTL has expired, so I think the comparison shouldn't be here. Also, I don't think an error should be logged since it's an expected flow.

internal/controller/clusterorder_controller.go
```diff
@@ -244,9 +248,9 @@ func (r *ClusterOrderReconciler) handleUpdate(ctx context.Context, _ ctrl.Reques
 	} else {
 		// only trigger webhook if the hostedcluster does not exist
 		if url := r.CreateClusterWebhook; url != "" {
-			if err := triggerWebHook(ctx, url, instance); err != nil {
+			if remainintTime, err := triggerWebHook(ctx, url, instance, r.MinimumRequestInterval); err != nil {
```
adriengentil, line 251:
```suggestion
			if remainingTime, err := triggerWebHook(ctx, url, instance, r.MinimumRequestInterval); err != nil {
```

internal/controller/clusterorder_webhook.go
jkary: I believe the cache will return an item object if found otherwise the err variable is set. [link](https://github.com/muesli/cache2go/blob/518229cd8021d8568e4c6c13743bb050dc1f3a05/cachetable.go#L303).

You are correct about the delta.  If we get a result then we must be inside the minimumRequestInterval.

Returning an error is also incorrect as well.  Thanks for that catch.

internal/controller/clusterorder_controller.go
```diff
@@ -343,10 +345,18 @@ func (r *ClusterOrderReconciler) handleNoHostedCluster(ctx context.Context,
...
 			log.Error(err, fmt.Sprintf("Failed to trigger webhook %s: %v", url, err))
 			return ctrl.Result{Requeue: true}, nil
 		}
+
+		// Verify if we are within the minimum request window
+		if remainingTime != 0 {
+			return ctrl.Result{RequeueAfter: remainingTime}, nil
+		}
+		log.Error(err, fmt.Sprintf("Failed to trigger webhook %s: %v", url, err))
+		return ctrl.Result{Requeue: true}, nil
```
adriengentil, line 359: how that requeue works internally? Asking because I would like to understand the impact of the requeue mechanism on the memory if we have a large among of items in it.

adriengentil (approved):
I just have a performance question but I think we can merge and see if it's an issue

internal/controller/clusterorder_controller.go
```diff
@@ -64,10 +64,11 @@ func (r *ClusterOrderReconciler) components() []component {
...
 	client.Client
-	Scheme                *runtime.Scheme
-	CreateClusterWebhook  string
-	DeleteClusterWebhook  string
-	ClusterOrderNamespace string
+	Scheme                 *runtime.Scheme
+	CreateClusterWebhook   string
+	DeleteClusterWebhook   string
+	ClusterOrderNamespace  string
+	MinimumRequestInterval string
```
larsks, line 71: Why isn't this a `time.Duration` variable (echoing @mhrivnak's comment elsewhere, why are we re-parsing this every time we call `triggerWebHook`)? And where does it get set? This structure is initialized by the `NewClusterOrderReconciler` method, but that method hasn't been updated to set this field.

larsks (changes requested):
Whether or not we make `triggerWebHook` a method on the controller (I think that sounds like a fine idea), I think we absolutely need to perform validation at startup and exit with an error if the value provided is invalid.

We also need to provide a mechanism for setting `MinimumRequestInterval`; a common model would be to add a command line option that takes a default from an environment variable.

internal/controller/clusterorder_controller.go
mhrivnak: It's a priority queue used by controller-runtime for a long time, and the same one where events from the k8s API server get queued in potentially high volume. This code path also includes automatic back-off. So I wouldn't worry too much about memory use here in general unless there's a specific concern.

internal/controller/clusterorder_webhook.go
```diff
@@ -8,38 +8,52 @@ import (
 	"net/http"
 	"time"
 
+	cache "github.com/muesli/cache2go"
```
mhrivnak, line 11: This appears to be a dead/abandoned project. Last release in 2018, and last code change in 2019 (not counting CI, documentation changes, etc, which have still been stale for 3+ years). Is there another option that's being actively maintained?

larsks: I haven't looked at it closely. but https://github.com/go-pkgz/expirable-cache seems to be maintained, and includes an interesting comparison with other caching module.

internal/controller/clusterorder_webhook.go
```diff
@@ -6,25 +6,71 @@ import (
...
-	cache "github.com/muesli/cache2go"
 	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
 
 	cloudkitv1alpha1 "github.com/innabox/cloudkit-operator/api/v1alpha1"
 )
 
-var inflightRequests *cache.CacheTable = cache.Cache("inflightRequests")
+type InflightRequest struct {
+	createTime             time.Time
+	minimumRequestInterval time.Duration
```
mhrivnak, line 19: What's the thinking around storing a copy of this with every request? AFAIK it's the same for every request and set by configuration.

jkary: So at a minimum we need a createTime timestamp.   The minReqInterval is able to be set for individual requests.  This was my interpretation of how the trigger should be called.  I can make it part of the configuration, however may I suggest I do this as a separate patch?

internal/controller/clusterorder_webhook.go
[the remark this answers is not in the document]
jkary: see [below](https://github.com/innabox/cloudkit-operator/pull/40#discussion_r2102678092)

internal/controller/clusterorder_controller.go
```diff
@@ -343,10 +345,18 @@ func (r *ClusterOrderReconciler) handleNoHostedCluster(ctx context.Context,
...
+		if err != nil {
 			log.Error(err, fmt.Sprintf("Failed to trigger webhook %s: %v", url, err))
 			return ctrl.Result{Requeue: true}, nil
 		}
+
+		// Verify if we are within the minimum request window
+		if remainingTime != 0 {
+			return ctrl.Result{RequeueAfter: remainingTime}, nil
+		}
+		log.Error(err, fmt.Sprintf("Failed to trigger webhook %s: %v", url, err))
```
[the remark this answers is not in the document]
jkary: fixed.

internal/controller/clusterorder_webhook.go
jkary: refactored into own caching mechanism using sync package.

internal/controller/clusterorder_webhook.go
```diff
@@ -6,40 +6,100 @@ import (
...
+		createTime:             time.Now(),
+		minimumRequestInterval: minimumRequestInterval,
+	}
+	ifRequestsLock.Unlock()
+	purgeExpiredRequests()
+}
+
+func purgeExpiredRequests() {
+	for url, request := range inflightRequests {
+		ifRequestsLock.RLock()
```
mhrivnak, line 55: Again, once you start iterating, you're already reading the map. Here it is too late to get the lock. You need the lock before the `range` statement.

By the time we get to this line, the first key and value have already been read from the map and copied into the values `url` and `request`.

mhrivnak: Maybe I commented too quickly; did you push a change yet? I saw that the above comment was resolved and assumed this was the current code. Maybe I should be more patient. :)

internal/controller/clusterorder_webhook.go
```diff
@@ -6,40 +6,102 @@ import (
...
+
+	inflightRequestsLock.RLock()
+	for url, request := range inflightRequests {
+		if delta := time.Since(request.createTime); delta > minimumRequestInterval {
+			expiredRequests = append(expiredRequests, url)
+		}
+	}
+	inflightRequestsLock.RUnlock()
+
+	if len(expiredRequests) > 0 {
```
mhrivnak, line 62: This will always be true, because you made the size 3 above where you initialized it.

internal/controller/clusterorder_webhook.go
```diff
@@ -6,40 +6,102 @@ import (
...
+	inflightRequestsLock.Lock()
+	inflightRequests[url] = InflightRequest{
+		createTime: time.Now(),
+	}
+	inflightRequestsLock.Unlock()
+	purgeExpiredRequests(minimumRequestInterval)
+}
+
+func purgeExpiredRequests(minimumRequestInterval time.Duration) {
+	var expiredRequests = make([]string, 3)
```
mhrivnak, line 52: Just initialize this empty. Otherwise you'll make it have a size of 3, and the first three items will each be an empty string (the "zero value" for type string).

```
expiredRequests := []string{}
```

internal/controller/clusterorder_webhook.go
```diff
@@ -6,40 +6,102 @@ import (
...
+	for url, request := range inflightRequests {
+		if delta := time.Since(request.createTime); delta > minimumRequestInterval {
+			expiredRequests = append(expiredRequests, url)
+		}
+	}
+	inflightRequestsLock.RUnlock()
+
+	if len(expiredRequests) > 0 {
+		inflightRequestsLock.Lock()
+		for url := range inflightRequests {
```
mhrivnak, line 64: This will give you indexes, not values. https://gobyexample.com/range

To ignore indexes but use values, you want:
```
for _, url := range inflightRequests {
```

mhrivnak (changes requested):
Getting close!

internal/controller/clusterorder_webhook.go
jkary: Of course!  Duh!

internal/controller/clusterorder_webhook.go
jkary: Sorry.  I’ve been staring at so many iterations that I’m missing little details

internal/controller/clusterorder_webhook.go
```diff
@@ -6,40 +6,102 @@ import (
...
 	defer resp.Body.Close() //nolint:errcheck
 
 	// Check response status
 	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
-		return fmt.Errorf("received non-success status code: %d", resp.StatusCode)
+		return 0, fmt.Errorf("received non-success status code: %d", resp.StatusCode)
 	}
 
-	return nil
+	addInflightRequest(url, minimumRequestInterval)
```
larsks, line 105: This never gets called -- we exit above for successful calls to the webhook, so you're never adding URLs to inflightRequests and no rate limiting is being applied.

larsks (changes requested):
I ran this in a test environment and it looks like the rate limiting never happens; I left a comment at the problematic spot.

internal/controller/clusterorder_webhook.go
larsks: I added some additional logging; it looks like urls are being purged from `inflightRequests` immediately after they are added.

cmd/main.go
```diff
@@ -208,12 +216,19 @@ func main() {
 		setupLog.Info("gRPC connection to fulfillment service is disabled")
 	}
 
+	interval, err := time.ParseDuration(minimumRequestInterval)
+	if err != nil {
+		setupLog.Error(err, "Invalid minimum request interval.")
+		os.Exit(1)
+	}
```
larsks, line 223: What is the behavior here if `minimumRequestInterval` is unset?

larsks:
How to test all this locally (I'm going to add something like this to the `README`, but I wanted to post this here so you would have it immediately):

1. Install [Docker](https://www.docker.com/) (it's possible to do this using Podman instead of Docker, but it's not quite as smooth).

1. Install [`kind`](https://kind.sigs.k8s.io/).

1. Create a new cluster:

    ```
    kind create cluster --kubeconfig kubeconfig --name innabox
    ```

    This creates a new Kubernetes cluster and puts administrative credentials in the file `kubeconfig` in your current directory.

1. Make the new `kubeconfig` file the default:

    ```
    export KUBECONFIG=$PWD/kubeconfig
    ```

1. Deploy the CRDs into the new cluster:

    ```
    make install
    kubectl apply -k config/crd/fakes
    ```

1. Create a namespace in which you will create ClusterOrders:

    ```
    kubectl create ns cloudkit-orders
    ```

1. Run a simple web service to provide an HTTP webhook receiver:

    ```
    podman run -d --name whoami -p 8080:80 docker.io/traefik/whoami
    ```

   (You can use `docker` instead of `podman` if that's what you have available locally.) 

1. Build the controller:

    ```
    make
    ```

1. Set environment variables as necessary:

    ```
    export CLOUDKIT_CLUSTER_CREATE_WEBHOOK=http://localhost:8080/create
    export CLOUDKIT_CLUSTER_DELETE_WEBHOOK=http://localhost:8080/delete
    export CLOUDKIT_CLUSTER_ORDER_NAMESPACE=cloudkit-orders
    ```

1. Run the controller:

    ```
    ./bin/manager
    ```

1. Deploy a sample cluster order:

    ```
    kubectl -n cloudkit-orders create -f config/samples/cloudkit_v1alpha1_clusterorder.yaml
    ```

Watch the logs from the `manager` process and ensure things behave as expected. You can trigger a reconcile on the ClusterOrder (which will in turn trigger a call to the webhook if one is defined) by making any change to the resource; for example:

```
kubectl patch clusterorder clusterorder-sample --type merge --patch "
apiVersion: cloudkit.openshift.io/v1alpha1
kind: ClusterOrder
metadata:
  annotations:
    force-refresh: '$RANDOM'
"
```

(This will add or update the `force-refresh` annotation with a random value.)

cmd/main.go
jkary: Uninitialized strings default to "" and time.ParseDuration("") yields an error.

internal/controller/clusterorder_webhook.go
larsks: Right. Pretty much the same code, but ranging over `expiredRequests` rather than `inflightRequests`.

cmd/main.go
larsks: Possibly by doing something as simple as:

```go
if minimumRequestInterval == "" {
  minimumRequestInterval = "0"
}
```

larsks (approved):
Thanks for taking the time to deal with all our comments! This is going to be a useful change; I think the issue with the postgres volume filling up this morning was due in large part to the frequency with which we're triggering the webhooks.

PR #41, opened 2025-05-13
Fake fulfilled status for cluster order

larsks:
We need cluster orders to reach the "fulfilled" state. We would normally
look at something like the `ClusterVersionAvailable` status on the
HostedCluster resource, but because of innabox/issues#112, that condition
is always False due to the failing cluster operators.

In this commit, we translate the `Available` status on the HostedCluster
resource to "fulfilled" status in the API.


tzumainn (approved):
It works!

PR #43, opened 2025-05-21
Grant controller permission to read node pools

jhernand:
In order to send the descriptions of node pools to the fulfillment service the controller needs permission to read the `NodePool` objects.


larsks:
Closing in favor of #44.

PR #52, opened 2025-05-27
Teach controller about cloudkit.openshift.io/management-state annotation

larsks:
This is allows us to have the controller completely ignore a ClusterOrder
(management-state: unmanaged) or to never trigger a webhook for a
ClusterOrder (management-state: manual).


tzumainn (approved):
Makes sense - thanks!

PR #54, opened 2025-05-29
Copy initial node sets from template to cluster

jhernand:
When a cluster object is created in the fulfillment API it should have the `spec.node_sets` field initialized from the `node_sets` field of the template that was used to create the cluster order. This patch adds that logic.


larsks (commented):
I'm a little unclear what's going on here -- having the fulfillment controller read template information from the fulfillment service only to write that information back to the fulfillment service to update a cluster seems like the wrong place for this.

I don't think this controller shouldn't be interacting with templates.

The clusterorder should have an initial `spec.nodeRequests` populated by the fulfillment service using the template metadata. We're not publishing this right now because when we implemented https://github.com/innabox/issues/issues/121 there was no API support for receiving the noderequest information; if that has changed we should update the publishing job to also push this information.

This controller should simply ensure that the cluster resource in the fulfillment service reflects the clusterorder.

larsks (approved):
Approving so we can do the demo, but I think we'll need to work on this.

PR #55, opened 2025-05-29
Ensure that sample manifest is valid

larsks:


tzumainn (approved):
Confirmed that this sample works!

PR #58, opened 2025-05-31
Make logging more consistent

larsks:
Use key/value pairs rather than `fmt.Sprintf()`, ensure logs all start with
lower case letter.


tzumainn (approved):
I'm probably not used to go, but it's bizarre to me that a key/value pair is apparently just consecutive comma-separated arguments

PR #59, opened 2025-06-02
Revert "Copy initial node sets from template to cluster"

larsks:
As discussed in the original pull request (#54), nothing in the operator
should be interacting with templates; that should all happen in the
fulfillment service before delivering a ClusterOrder to the operator.

This reverts commit 3908f826975086051780d8a1511cfbf6e657b672.


tzumainn:
Just to clarify - does this reversion break any functionality that currently works? If so, is there an agreed-upon plan regarding how to restore that functionality? (I took a look at the PR you mentioned, but wasn't sure if there was a consensus.)

jhernand:
This has been addressed as part of the merge of `ClusterOrder` and `Cluster`. The only remaining thing to do is remove the unused templates client from the feedback controller, and that is addressed in #61, so I think this can be closed now.

PR #63, opened 2025-08-18
Ignore changes to resources managed by other operator instances

larsks:
Now that we are running multiple instance of the operator in a cluster, we
need to filter out change notifications for resources that are managed by
other instances.

This PR includes two changes:

- Update the feedback controller to only monitor clusterorder resources in a single namespace
- Ignore notifications from namespaces, etc, managed by other operator instances

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -204,6 +204,7 @@ func main() {
 			ctrl.Log.WithName("feedback"),
 			mgr.GetClient(),
 			grpcConn,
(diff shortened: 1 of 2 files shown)


jkary:
I don’t think we can assume this is the case, but it may be a version 2.0 goal?

I didn’t quite follow the caching conversation.  Why can’t we dynamically watch namespaces adding and deleting as necessary?

Jason

PR #69, opened 2025-09-23
Fix: Added check for VM != Ready before calling the webhook.

jkary:


adriengentil:
Can you explain why? Before reaching this block of code, we check if we can find a kubevirt VM, and then we return. Isn't it enough?

adriengentil:
>My understanding is cloudkit-aap will set the phase to Ready
it the operator that sets the phase to Ready here: https://github.com/innabox/cloudkit-operator/blob/main/internal/controller/virtualmachine_controller.go#L424

adriengentil (approved):

internal/controller/virtualmachine_controller.go
```diff
@@ -259,6 +259,10 @@ func (r *VirtualMachineReconciler) handleUpdate(ctx context.Context, _ ctrl.Requ
 		}
 	}
 
+	if instance.Status.Phase != v1alpha1.VirtualMachinePhaseReady {
```
adriengentil, line 262:
```suggestion
	if instance.Status.Phase == v1alpha1.VirtualMachinePhaseReady {
```

approved too fast

adriengentil (approved):

PR #159, opened 2026-03-25
MGMT-23662: persist subnet-namespace annotation and refactor annotation handling

eranco74:
## Summary
- Resolve subnet namespace from Subnet CR and persist it as the `osac.openshift.io/subnet-namespace` annotation so downstream reconciliation uses the correct VM search namespace
- Extract `syncSubnetNamespaceAnnotation` into its own method and define `osacSubnetNamespaceAnnotation` in `computeinstance_names.go` for consistency with other annotation variables
- Since `SubnetRef` is immutable, the annotation is resolved once and cached — subsequent reconciles reuse the persisted value without redundant API calls
- Only swallow subnet-not-found errors with a fixed 30s requeue; propagate infrastructure errors (`r.Update`/`r.Get` failures) so controller-runtime applies exponential backoff

## Test plan
- [x] Annotation persistence: reconcile with valid SubnetRef sets annotation on API server
- [x] No-op optimization: second reconcile with unchanged annotation skips `r.Update()` (verified via Generation + Annotations equality)
- [x] Missing Subnet CR: reconcile requeues after 30s without error
- [x] Empty SubnetRef: no subnet-namespace annotation is set
- [x] Existing SubnetRef with Subnet CR: reconcile succeeds with Starting phase

## Summary by CodeRabbit

* **Improvements**
  * Controller now gracefully requeues for 30s when a referenced Subnet is temporarily missing, avoiding failures.
  * ComputeInstance metadata annotation for subnet-namespace is persisted idempotently and reused to reduce redundant lookups.
  * Status updates now use the server-fetched resource after changes to ensure consistent, server-populated fields.

### DIFF
--- a/internal/controller/computeinstance_controller.go
+++ b/internal/controller/computeinstance_controller.go
@@ -53,6 +53,11 @@ const (
 	DefaultStatusPollInterval = 30 * time.Second
 )
 
+// errSubnetNotFound is returned when the Subnet CR referenced by SubnetRef
+// does not exist. handleUpdate treats this as a transient error and requeues
+// with a fixed delay instead of exponential backoff.
+var errSubnetNotFound = errors.New("subnet CR not found")
+
 // ComputeInstanceReconciler reconciles a ComputeInstance object
 type ComputeInstanceReconciler struct {
 	client.Client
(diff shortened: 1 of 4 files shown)


tzvatot (approved):
/lgtm

PR #180, opened 2026-04-14
MGMT-23859: fix subnet feedback controller stuck on NotFound during deletion

omer-vishlitzky:
https://redhat.atlassian.net/browse/MGMT-23859

The subnet feedback controller fetches the subnet record from the fulfillment API before checking deletion state. During deletion the record is already archived, causing NotFound errors that prevent finalizer removal. The CR gets stuck in Deleting forever with only the subnet-feedback finalizer remaining. Handle NotFound when the CR is being deleted by removing the finalizer and returning.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced resource deletion handling to gracefully manage cases where backend records are unavailable, allowing proper cleanup without error propagation
  
* **Tests**
  * Added test cases verifying correct deletion behavior when backend records are missing

### DIFF
--- a/internal/controller/subnet_feedback_controller.go
+++ b/internal/controller/subnet_feedback_controller.go
@@ -17,6 +17,8 @@ import (
 	"context"
 
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	clnt "sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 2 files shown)


internal/controller/subnet_feedback_controller.go
```diff
@@ -93,6 +95,13 @@ func (r *SubnetFeedbackReconciler) Reconcile(ctx context.Context, request ctrl.R
 	// Step 3: Fetch the subnet from the fulfillment service so we can compare before/after.
 	subnet, err := r.fetchSubnet(ctx, subnetID)
 	if err != nil {
+		if !object.DeletionTimestamp.IsZero() && status.Code(err) == codes.NotFound {
```
eranco74, line 98: The fix is valid, unsure how it got to a state where the subnet is archived in the fulfilment service but still exists as k8s CR on the hub??

eranco74 (commented):
/lgtm

internal/controller/subnet_feedback_controller.go
akshaynadkarni: +1 

Also, does a similar issue exist for `ComputeInstance` and other resources?

PR #182, opened 2026-04-14
Remove deprecated host class API files

tzumainn:
These files are now deprecated with the renaming of host class
to host type.


larsks (commented):
This says, "Rename HostClass to HostType", but it only contains delete operations for `host_class*` files. It does not add any replacements. Is the title erroneous or is the PR missing something?

tzumainn:
> This says, "Rename HostClass to HostType", but it only contains delete operations for `host_class*` files. It does not add any replacements. Is the title erroneous or is the PR missing something?

Ah - turns out someone beat me to the renaming: https://github.com/osac-project/osac-operator/commit/4ad2b2da0cdcdca69e4a9ed47e9f92f6267e4a2c

I've retitled the commit to specify that all it does is remove the deprecated host class API files

adriengentil (approved):

PR #187, opened 2026-04-16
MGMT-23392: Add VMaaS dedicated cluster documentation

adriengentil:
## Summary

- Add `docs/vmaas-dedicated-cluster/README.md` documenting the multi-cluster feature
- Covers motivation (Enclave use case), architecture, configuration for both osac-operator and AAP, setup steps, constraints, and references to the originating PRs and ticket

## References

- [MGMT-23392](https://redhat.atlassian.net/browse/MGMT-23392)
- [MGMT-23102](https://redhat.atlassian.net/browse/MGMT-23102)
- osac-project/osac-operator#119
- osac-project/osac-aap#209
- osac-project/osac-aap#213

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive guide for running VM management against a separate remote VMaaS cluster, including architecture, setup steps, operator and orchestration configuration for using a remote kubeconfig, instructions for provisioning the kubeconfig secret, and example deployment snippets.
  * Clarifies supported controllers and compatibility constraints when using remote mode.

### DIFF
--- a/docs/vmaas-dedicated-cluster/README.md
+++ b/docs/vmaas-dedicated-cluster/README.md
@@ -1,0 +1,151 @@
+# VMaaS Dedicated Cluster
+
+The OSAC operator can manage KubeVirt VMs on a **remote cluster** that is
+separate from the cluster where the operator itself runs. This allows all VMs
+to be deployed to a dedicated VMaaS cluster while the operator, fulfillment
+service, and AAP remain on the management cluster.
+
+## Motivation
+
+This feature was originally designed for the
+[Enclave](https://github.com/rh-ecosystem-edge/enclave) use case, where the
+control plane (fulfillment service, osac-operator, and AAP) and the VMaaS
+cluster are managed by different personas.
+
+## Architecture
+
+| Controller | Local cluster | VMaaS cluster (remote) |
+|---|---|---|
+| ComputeInstance controller | reconciles ComputeInstance CRs | watch KubeVirt VirtualMachine/VirtualMachineInstance |
+| Tenant controller | reconciles Tenant CRs | manage Tenant namespace, UDN resources |
+| AAP | | manage KubeVirt VirtualMachine/VirtualMachineInstance |
+
+The `ComputeInstance` and `Tenant` custom resources remain on the management
+cluster. Their controllers reconcile those resources locally and watch the
+downstream Kubernetes objects (VMs, namespaces, UDN) on the remote cluster
+via a dedicated kubeconfig.
+
+osac-operator relies on
+[multicluster-runtime](https://github.com/kubernetes-sigs/multicluster-runtime)
+to watch and manage resources across both clusters. AAP relies on the
+standard [kubernetes.core](https://github.com/ansible-collections/kubernetes.core)
+Ansible collection, passing a kubeconfig to each task to target the remote
+cluster.
+
+The `ClusterOrder` and `HostPool` controllers are local-only and are
+incompatible with the remote cluster option (see [Constraints](#constraints)).
+
+## Configuration
+
+### osac-operator
+
+The operator accepts a kubeconfig path that points to the remote cluster. When
+set, the `ComputeInstance` and `Tenant` controllers target the remote cluster
+instead of the local one.
+
+| Flag | Environment variable | Default | Description |
+|---|---|---|---|
+| `--remote-cluster-kubeconfig` | `OSAC_REMOTE_CLUSTER_KUBECONFIG` | _(unset)_ | Path to a kubeconfig file for the remote VMaaS cluster. When unset the operator behaves as before (single-cluster mode). |
+| `--enable-tenant-controller` | `OSAC_ENABLE_TENANT_CONTROLLER` | `false` | Enable the Tenant controller. |
+| `--enable-compute-instance-controller` | `OSAC_ENABLE_COMPUTE_INSTANCE_CONTROLLER` | `false` | Enable the ComputeInstance controller. |
+
+> **Note:** When no `--enable-*-controller` flag is set, the operator enables
+> all controllers by default. When using a remote cluster, at least one of
+> `--enable-tenant-controller` or `--enable-compute-instance-controller` must
+> be set explicitly to avoid also enabling the controllers that are
+> incompatible with the remote cluster option.
+
+The kubeconfig is typically provided to the operator pod as a mounted
+Kubernetes Secret volume.
+
+#### Example: operator deployment with remote cluster
+
+```yaml
+env:
+  - name: OSAC_REMOTE_CLUSTER_KUBECONFIG
+    value: /etc/osac/remote-cluster-kubeconfig
+  - name: OSAC_ENABLE_TENANT_CONTROLLER
+    value: "true"
+  - name: OSAC_ENABLE_COMPUTE_INSTANCE_CONTROLLER
+    value: "true"
+volumeMounts:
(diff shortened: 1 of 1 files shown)


docs/vmaas-dedicated-cluster/README.md
```diff
@@ -0,0 +1,150 @@
+# VMaaS Dedicated Cluster
+
+The OSAC operator can manage KubeVirt VMs on a **remote cluster** that is
+separate from the cluster where the operator itself runs. This allows all VMs
+to be deployed to a dedicated VMaaS cluster while the operator, fulfillment
```
akshaynadkarni, line 5: I think fulfillment service will soon be renamed to osac service.

docs/vmaas-dedicated-cluster/README.md
```diff
@@ -0,0 +1,150 @@
...
+## Motivation
+
+This feature was originally designed for the
+[Enclave](https://github.com/rh-ecosystem-edge/enclave) use case, where the
+control plane (fulfillment service, osac-operator, and AAP) and the VMaaS
+cluster are managed by different personas.
+
+## Architecture
+
+| Controller | Local cluster | VMaaS cluster (remote) |
```
akshaynadkarni: Should we say "Management Cluster" instead of "Local cluster"?

docs/vmaas-dedicated-cluster/README.md
```diff
@@ -0,0 +1,150 @@
...
+|---|---|---|
+| ComputeInstance controller | reconciles ComputeInstance CRs | watch KubeVirt VirtualMachine/VirtualMachineInstance |
+| Tenant controller | reconciles Tenant CRs | manage Tenant namespace, UDN resources |
+| AAP | | manage KubeVirt VirtualMachine/VirtualMachineInstance |
```
akshaynadkarni: Question: Does AAP have to be installed on the management cluster or just the VMaaS cluster? In my setup, they both are the same.

adriengentil: on the management cluster as stated in the introduction: "This allows all VMs
to be deployed to a dedicated VMaaS cluster while the operator, fulfillment 
service, and AAP remain on the management cluster."

docs/vmaas-dedicated-cluster/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+### osac-operator
+
+The operator accepts a kubeconfig path that points to the remote cluster. When
+set, the `ComputeInstance` and `Tenant` controllers target the remote cluster
+instead of the local one.
+
+| Flag | Environment variable | Default | Description |
+|---|---|---|---|
+| `--remote-cluster-kubeconfig` | `OSAC_REMOTE_CLUSTER_KUBECONFIG` | _(unset)_ | Path to a kubeconfig file for the remote VMaaS cluster. When unset the operator behaves as before (single-cluster mode). |
```
akshaynadkarni: (minor) I don't think we should say "as before", since someone interacting with our repo for the first time may not be aware of this behavior.

```suggestion
| `--remote-cluster-kubeconfig` | `OSAC_REMOTE_CLUSTER_KUBECONFIG` | _(unset)_ | Path to a kubeconfig file for the remote VMaaS cluster. When unset the operator behaves in single-cluster mode. |
```

docs/vmaas-dedicated-cluster/README.md
adriengentil: changed it, thanks!

docs/vmaas-dedicated-cluster/README.md
```diff
@@ -0,0 +1,150 @@
...
+| `--enable-tenant-controller` | `OSAC_ENABLE_TENANT_CONTROLLER` | `false` | Enable the Tenant controller. |
+| `--enable-compute-instance-controller` | `OSAC_ENABLE_COMPUTE_INSTANCE_CONTROLLER` | `false` | Enable the ComputeInstance controller. |
+
+> **Note:** When no `--enable-*-controller` flag is set, the operator enables
```
akshaynadkarni, line 52: (question) 
- Which controllers are incompatible with the remote cluster option? 
- Why do we require these to be set to `true` explicitly for the remote cluster config?

adriengentil: >Which controllers are incompatible with the remote cluster option?

all other ones, I don't want to mention them because the code changes quickly

>Why do we require these to be set to true explicitly for the remote cluster config?

because all controllers are enabled by default, see the note below

docs/vmaas-dedicated-cluster/README.md
adriengentil: ok let's rename now then

docs/vmaas-dedicated-cluster/README.md
```diff
@@ -0,0 +1,150 @@
...
+  --from-file=kubeconfig=/path/to/remote-cluster.kubeconfig \
+  -n <osac-namespace>
+```
+
+### 2. Configure the operator
+
+Set `OSAC_REMOTE_CLUSTER_KUBECONFIG` and mount the Secret (see example
+above). Enable only the controllers that should run against the remote cluster
+(`tenant` and/or `compute-instance`). Other controllers, if needed, must run
+in a separate operator instance without the remote kubeconfig.
```
akshaynadkarni, line 124: (just curious) 
Now we have virtual network, subnet etc. controllers. When you say those should run in a separate operator instance, do you mean there will be 2 instances of the osac-operator, one with the "tenant and computeinstance controller" and the other with "virtual network, subnet etc. controllers" ?

It seems like a requirement of the remote cluster kubeconfig is for the "tenant and compute instance" controllers to run in a separate operator instance. Is that right? 

If yes, could we automatically do that once remote cluster kubeconfig is set? What's the use-case for the additional enable flags?

docs/vmaas-dedicated-cluster/README.md
akshaynadkarni: Ah yes 🤦‍♂️ . 
Then maybe for the middle column you want to say "executes playbooks" or something similar, just to make it clear that AAP is running on management cluster.

docs/vmaas-dedicated-cluster/README.md
adriengentil: >  Now we have virtual network, subnet etc. controllers. When you say those should run in a separate operator instance, do you mean there will be 2 instances of the osac-operator, one with the "tenant and computeinstance controller" and the other with "virtual network, subnet etc. controllers" ?

As it is implemented today, yes.

Networking is a bad example, because it needs to support VM networking, and eventually will be remote-aware (we have a ticket about it). Cluster-as-a-service is a better example, as we way be in a setup where VMs are spawn on a remote cluster, but hosted clusters are created on the management cluster.

The idea of this remote flag, is to turn the operator in a remote-mode as a whole. So, it means each controller must be aware of this mode, and act accordingly. This might not be the best in terms of flexibility, but it was a straight way to deliver such mode in the short time we had to deliver this feature. Of course it can be re-vamped in the future, when requirements will be clearer.

> If yes, could we automatically do that once remote cluster kubeconfig is set?

We can change the behavior of the flags if we want to make things easier, the goal of this doc is mainly to state the existing, then we can move forward.

> What's the use-case for the additional enable flags?

For exampe in VMaaS deployment only, we don't want to register HCP CRD from the CaaS for example, because they won't be available. These flags give flexibility is term of deployment.

akshaynadkarni (approved):
LGTM

PR #185, opened 2026-04-16
MGMT-23607: add CLAUDE.md documentation for AI-assisted development

eranco74:
## Summary

Add comprehensive CLAUDE.md documentation to guide AI-assisted development in the osac-operator repository.

The documentation covers:
- 📋 Project overview and managed resources (ClusterOrder, ComputeInstance, networking resources)
- 🛠️ Development commands (build, test, deploy, code generation)
- 🏗️ Architecture patterns (dual-controller pattern, provisioning providers, multi-cluster support)
- 📝 CRD development patterns with kubebuilder markers
- 🔄 Controller reconciliation and finalizer management patterns
- 🧪 Testing strategies (unit, integration, cluster testing)
- ⚠️ Common pitfalls and solutions
- 🔗 Cross-repo coordination workflows

Also removes CLAUDE.md from .gitignore to allow committing the file.

## Test plan

- [x] File follows the established CLAUDE.md pattern from osac-aap
- [x] Documentation is comprehensive and covers all major components
- [x] Code examples are syntactically correct
- [x] Links to related documentation are valid

## Related Issues

- Jira: [MGMT-23607](https://issues.redhat.com/browse/MGMT-23607)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive documentation covering operator architecture, development workflows, configuration, and testing practices.

PR #196, opened 2026-04-23
MGMT-23607: Add CLAUDE.md with modular .claude/rules/

eranco74:
## Summary

Supersedes #185 — same CLAUDE.md content, restructured for best practices.

- Reduced `CLAUDE.md` from **802 lines to 125 lines** (84% reduction)
- Moved detailed reference content into 4 modular `.claude/rules/` files (224 lines total):
  - `controller-patterns.md` — Dual-controller, reconciliation, finalizer, AAP, feedback, CRD type patterns
  - `common-pitfalls.md` — 10 common issues with solutions
  - `common-tasks.md` — Adding CRDs/fields, cross-repo change order, RBAC, debugging
  - `configuration.md` — Environment variables for AAP, EDA, gRPC, namespaces, controller flags
- Root file keeps critical rules, build commands, architecture overview, and PR checklist
- Fixes MD040 markdownlint warnings (unlabeled fenced code blocks) flagged by CodeRabbit

## Why

[Claude Code best practices](https://docs.anthropic.com/en/docs/claude-code/memory#claude-md) recommend keeping CLAUDE.md under 200-300 lines. The root file is loaded into every conversation — 802 lines causes token bloat and reduces adherence. Rules in `.claude/rules/` are auto-loaded on demand.

## Test plan

- [ ] Verify `CLAUDE.md` renders correctly and covers all critical rules
- [ ] Verify `.claude/rules/` files are tracked in git
- [ ] Start a new Claude Code session and confirm rules are discoverable
- [ ] No markdownlint warnings on fenced code blocks

## Related Issues

- Jira: [MGMT-23607](https://issues.redhat.com/browse/MGMT-23607)
- Supersedes: #185

### DIFF
--- a/.claude/rules/common-pitfalls.md
+++ b/.claude/rules/common-pitfalls.md
@@ -1,0 +1,41 @@
+# Common Pitfalls
+
+## 1. Forgetting to regenerate after CRD changes
+Always run `make manifests generate` after modifying `api/v1alpha1/*.go`.
+CRD YAML in `config/crd/` and DeepCopy in `zz_generated.deepcopy.go` are generated — never edit directly.
+
+## 2. Status update loops
+Save `oldStatus := subnet.Status.DeepCopy()` before reconciliation.
+Use `equality.Semantic.DeepEqual` to compare. Only call `r.Status().Update()` if changed.
+
+## 3. Finalizer removal timing
+Remove finalizer only after cleanup fully completes. For feedback controllers: after syncing deletion state.
+Handle errors during cleanup — retry on next reconcile, don't remove prematurely.
+
+## 4. AAP job polling inefficiency
+Use `StatusPollInterval` (default: 30s). Only poll active jobs.
+Return `ctrl.Result{RequeueAfter: interval}` for delayed reconciliation.
+
+## 5. NotFound errors during deletion
+Feedback controllers may see NotFound if fulfillment-service archives before finalizer removed.
+Handle gracefully: check `DeletionTimestamp` + `codes.NotFound`, then remove finalizer.
+
+## 6. Parent resource lookup failures
+If parent not found (e.g., Subnet → VirtualNetwork), requeue with delay.
+Don't set Phase to Failed — parent may be created soon. Use conditions for transient errors.
+
+## 7. Vendor directory confusion
+Operator uses Go modules, not vendoring. Delete `vendor/` if it exists. Use `go mod tidy`.
+
+## 8. Integration test failures
+- Update `kustomization.yaml` after renaming/deleting manifests
+- Clean up test clusters: `kind delete cluster --name osac`
+- Run `make test-kustomize` before committing manifest changes
+
+## 9. gRPC client version mismatches
+Update module version in `buf.gen.yaml`, never edit proto files directly.
+Run `buf generate` to regenerate `internal/api/`. Commit generated files.
+
+## 10. AAP template naming
+Template names: `{prefix}-{action}-{resource-kind}` (e.g., `osac-create-subnet`).
+Per-resource overrides (e.g., `OSAC_CLUSTER_AAP_PROVISION_TEMPLATE`) take precedence.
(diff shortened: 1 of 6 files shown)


tzvatot (commented):
## Review: MGMT-23607 — Add CLAUDE.md with modular .claude/rules/

Good structure — 802→125 line reduction in root file is the right call, and the `.claude/rules/` modular approach will scale well. Content is accurate and well-organized.

I went through the git history and PR review comments to find lessons learned that could strengthen the rules files. These are real bugs and review findings that cost debugging time — exactly the kind of institutional knowledge that should be codified.

| Category | Count |
|----------|-------|
| 🔴 Critical | 0 |
| 🟡 Important | 1 |
| 💡 Suggestion | 0 |

---

### 🟡 Rules files are missing lessons learned from git history

The current rules capture general patterns well, but miss several hard-won lessons from actual bugs and review feedback. I'd suggest adding these to the existing rules files:

**Add to `common-pitfalls.md`:**

```markdown
## 11. Duplicate provision job triggers from stale cache
Use `r.APIReader` (uncached read from API server), not `r.Client` (cached), when
checking whether a non-terminal provision job already exists before triggering a new one.
The informer cache may be stale, causing duplicate AAP job triggers.
(PR #193, #131, #171)

## 12. Unconditionally setting Phase to Progressing
Never set phase to `Progressing` at the top of every reconcile. Set it only when:
(a) first provision (phase is empty), or (b) spec changed after a previous success.
During backoff, preserve `Failed` so users see the resource is in a failed-but-retrying state.
(PR #171, #130)

## 13. Feedback controller stuck on NotFound during deletion
In feedback controllers, handle `NotFound` from the fulfillment API when the CR is being
deleted (`DeletionTimestamp` is set). Remove the finalizer and return — the fulfillment
service may have already archived the record. Without this, the CR gets stuck in Deleting.
(PR #180)

## 14. Lost ErrorDetails in provision/deprovision updates
When updating provision or deprovision job status, always use `status.MessageWithDetails()`
to preserve ErrorDetails. All controller poll and trigger paths must propagate ErrorDetails.
(PR #158)

## 15. OnTriggered fires on rate-limit skip
Guard `OnTriggered` callbacks by comparing the latest job ID before and after `TriggerJob`,
not slice length (which fails when job history is trimmed). `TriggerJob` returns
`(RequeueAfter, nil)` for rate-limit responses without creating a job.
(PR #193)

## 16. Kubebuilder validation on status fields
Do not add `+kubebuilder:validation:Required` or similar markers to status subresource
fields. Status is set by the controller, not API consumers. Validation on status can
prevent status from being written at all.
(PR #154)

## 17. Missing scheme registration for multi-cluster
When adding a new controller or watching a new resource type, register required schemes
(e.g., `corev1`) in both the local and remote cluster managers. Also add all required
environment variables to `config/manager/manager.yaml` or the controller will silently
never reconcile.
(PR #132, #123)
```

**Add to `controller-patterns.md` under a new section:**

```markdown
## Concurrency in main.go

Use `errgroup` for managing concurrent blocking operations (remote cluster, remote
provider, manager). Never call `os.Exit()` from a goroutine — all errors must propagate
through the errgroup.
(PR #156)

## Shared Provisioning Logic

Provisioning and deprovisioning logic belongs in `internal/provisioning/`, not in
individual controllers or `internal/helpers/`. Avoid stutter in exported names
(use `provisioning.Backoff` not `provisioning.ProvisionBackoff`).
(PR #153, #158, #166)

## Spec vs Status Field Placement

Avoid spec fields that are convenience copies of data available through other paths
(annotations, parent resources, the full CR passed to Ansible). If the playbook receives
the full CR, do not also pass individual fields as extra vars.
(PR #142, #152)
```

**Add to `common-tasks.md` under Testing:**

```markdown
## Testing Conventions

- Use Ginkgo/Gomega for all controller tests — not standard `testing.T` (PR #156)
- Use `DeferCleanup()` from Ginkgo for resource cleanup (PR #155)
- Reference production constants in assertions instead of hardcoding values (PR #155)
```

tzvatot:
Re my review, I can push it in a followup PR.

/lgtm

PR #197, opened 2026-04-23
MGMT-23976: Add multi-tier storage types to Tenant CRD

zszabo-rh:
## Summary

- Add `ResolvedStorageClass` struct with `name` and `tier` fields to support multiple storage tiers per tenant (EP [#32](https://github.com/osac-project/enhancement-proposals/pull/32))
- Replace singular `status.storageClass` (string) with `status.storageClasses` (list of `ResolvedStorageClass`)
- Add `osac.openshift.io/storage-tier` label constant
- Update printer column from single StorageClass name to resolved tier names

## Notes

The controller is minimally adapted to write the existing single-tier result as a one-element list with `tier: "default"`. Full multi-tier resolution logic will follow in a separate PR once EP #32 is finalized.

## Test plan

- [x] `make test` passes — existing tenant controller tests adapted for the new list format
- [x] `make manifests` regenerates CRD YAML with correct schema

Jira: [MGMT-23976](https://redhat.atlassian.net/browse/MGMT-23976)

## Summary by CodeRabbit

* **Enhancements**
  * Tenant status now tracks multiple storage classes with tier information. The status field has been updated to display an array of storage class entries instead of a single value, each containing name and tier details.
  * Kubebuilder printer columns now display storage tiers for improved visibility.


akshaynadkarni (commented):
*(Reposted as inline comment on lines 385-392)*

internal/controller/tenant_controller.go
```diff
@@ -379,11 +381,13 @@ func (r *TenantReconciler) SetupWithManager(mgr mcmanager.Manager) error {
...
 func storageClassTenantPredicate() predicate.Predicate {
 	return predicate.NewPredicateFuncs(func(obj client.Object) bool {
-		_, exists := obj.GetLabels()[osacTenantAnnotation]
-		return exists
+		labels := obj.GetLabels()
+		_, hasTenant := labels[osacTenantAnnotation]
+		_, hasTier := labels[osacStorageTierLabel]
+		return hasTenant && hasTier
 	})
 }
```
akshaynadkarni, line 392: The predicate requires both labels, but `getTenantStorageClass` (line 206) still queries by tenant label only. This is intentionally ahead of the resolution logic until the full multi-tier work lands (EP #32). Consider adding a brief code comment noting that StorageClasses without the `storage-tier` label won't trigger reconciliation, so developers working on this area understand the mismatch is intentional while multi-tier resolution is in progress.

akshaynadkarni (approved):
Left a minor comment. You can address in a follow-up PR in case it will be a while until the watch and reconciliation logic matches. 

LGTM

PR #199, opened 2026-04-24
MGMT-23977: Implement per-tier StorageClass resolution in Tenant controller

zszabo-rh:
## Summary

- Replace single-tier `getTenantStorageClass()` with `getTenantStorageClasses()` that resolves all storage tiers independently
- For each distinct `storage-tier` label value, apply two-step fallback: tenant-specific SC first, then shared Default SC
- StorageClasses without the `osac.openshift.io/storage-tier` label are ignored
- Tier values normalized to lowercase
- Tenant is `Ready` when at least one tier resolves; `Progressing` when none do
- Per-tier duplicate detection emits `DuplicateStorageClass` warning events without blocking other tiers
- Emit `StorageClassNotReady` warning event when no tier resolves
- Adds `groupByTier()` helper that classifies StorageClasses by tier label
- Deprecated `status.storageClass` field populated from the first resolved tier for backward compatibility ([MGMT-24139](https://redhat.atlassian.net/browse/MGMT-24139))

Implements the resolution algorithm from EP [#32](https://github.com/osac-project/enhancement-proposals/pull/32).

## Test plan

- [x] `make test` passes — 12-step integration test covering: no namespace, no SCs, SC without tier label ignored, shared Default fallback, tenant-specific priority, multi-tier resolution, per-tier duplicate isolation, mixed resolution, full degradation with NotReady event, all-tiers-duplicate
- [x] Unit tests for `groupByTier()` (including lowercase normalization) and `joinStorageClassNames()`

Jira: [MGMT-23977](https://redhat.atlassian.net/browse/MGMT-23977)

## Summary by CodeRabbit

* **New Features**
  * Storage is resolved per storage-tier; tenant status lists all resolved storage classes by tier and still marks a primary choice.
  * Emits warning events when tiers are unresolvable; new distinct event reason for "not ready" cases.

* **Chores**
  * Simplified tenant condition reasons (removed redundant constants).

* **Tests**
  * Expanded coverage for multi-tier resolution, grouping/normalization, per-tier priority/fallback, and duplicate/misconfiguration cases.

### DIFF
--- a/internal/controller/tenant_controller.go
+++ b/internal/controller/tenant_controller.go
@@ -19,6 +19,7 @@ package controller
 import (
 	"context"
 	"fmt"
+	"sort"
 	"strings"
 
 	ovnv1 "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/userdefinednetwork/v1"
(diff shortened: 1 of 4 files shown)


internal/controller/tenant_controller.go
akshaynadkarni: I guess you should remove this for the CI to pass.

internal/controller/tenant_controller_test.go
```diff
@@ -105,13 +117,10 @@ var _ = Describe("Tenant Controller", func() {
 					_ = doReconcile()
 					g.Expect(k8sClient.Get(ctx, typeNamespacedName, tenant)).To(Succeed())
 					g.Expect(tenant.Status.Phase).To(Equal(expectedPhase))
-					g.Expect(tenant.Status.StorageClass).To(Equal(expectedSC))
-					if expectedSC == "" {
+					if expectedSCs == nil {
```
akshaynadkarni, line 120: Should we check `status.StorageClass` here too? It's [populated for backward compatibility](https://github.com/osac-project/osac-operator/blob/34f43b8f04291faa0c1a8811312dac7db2f62417/internal/controller/tenant_controller.go#L178) but not asserted anywhere. Something like:

```go
if expectedSCs == nil {
    g.Expect(tenant.Status.StorageClass).To(BeEmpty())
    g.Expect(tenant.Status.StorageClasses).To(BeNil())
} else {
    g.Expect(tenant.Status.StorageClasses).To(ConsistOf(expectedSCs))
    g.Expect(tenant.Status.StorageClass).To(Equal(expectedSCs[0].Name))
}
```

akshaynadkarni:
`TenantReasonSharedDefault` and `TenantReasonMultipleDefaultsFound` ([tenant_types.go#L54-L56](https://github.com/osac-project/osac-operator/blob/34f43b8f04291faa0c1a8811312dac7db2f62417/api/v1alpha1/tenant_types.go#L54-L56)) are no longer set by the controller after the per-tier refactor. Since this is pre-release v1alpha1, should we remove them to avoid dead code?

internal/controller/tenant_controller.go
```diff
@@ -144,46 +145,56 @@ func (r *TenantReconciler) handleUpdate(ctx context.Context, req reconcile.Reque
...
 		instance.SetStatusCondition(v1alpha1.TenantConditionStorageClassReady,
 			metav1.ConditionFalse,
-			scResult.reason,
-			scResult.message)
-		if scResult.reason == v1alpha1.TenantReasonMultipleFound || scResult.reason == v1alpha1.TenantReasonMultipleDefaultsFound {
-			r.Recorder.Eventf(instance, nil, corev1.EventTypeWarning, eventReasonDuplicateStorageClass, eventActionDetectDuplicate, "%s", scResult.message)
-		}
+			reason,
+			condMsg)
+		r.Recorder.Eventf(instance, nil, corev1.EventTypeWarning, eventReasonStorageClassNotReady, "StorageClassResolution", "%s", condMsg)
```
akshaynadkarni, line 169: **Question: When should a Tenant be considered not ready due to StorageClass issues?**

Currently `StorageClassNotReady` is emitted only when zero tiers resolve. If some tiers resolve and others have duplicates, the Tenant is Ready and the only signal is a `DuplicateStorageClass` event.

This raises a question: should a Tenant be Ready when a tier that was explicitly configured for it doesn't resolve? For example, if an admin creates a tenant-specific "fast" StorageClass, then accidentally introduces a duplicate, the Tenant stays Ready. The misconfiguration is only visible in events, not in the phase.

Before deciding on the event model (`StorageClassNotReady` vs `TenantNotReady` vs `TenantTierNotReady`), it's worth clarifying the readiness semantics:

**Option 1: At least one tier resolves = Ready (current, per EP #32)**
- `StorageClassNotReady` only fires when all tiers fail
- Pro: lenient, unrelated tier misconfigurations don't block the tenant
- Con: hides misconfigurations in tiers explicitly set up for the tenant

**Option 2: All discovered tiers must resolve = Ready**
- Any tier failure blocks readiness
- Pro: catches all misconfigurations
- Con: a duplicate in a shared Default tier the tenant doesn't need would block it

**Option 3: Spec-driven required tiers**
- Add `requiredTiers` to TenantSpec, controller checks exactly those
- Pro: explicit contract, cleanest model
- Con: CRD change, broader scope

This PR implements Option 1 per the EP. Raising it here since the answer shapes what events we emit and when. Happy to discuss offline.

zszabo-rh: I think option 1 is the right starting point for v1, a misconfiguration in one tier shouldn't block the entire tenant when other tiers are fine. 

The DuplicateStorageClass event already signals per-tier misconfigurations. If we find that operators need stronger guarantees, option 3 would be the cleanest future evolution, it makes the contract explicit without overloading the resolution logic.

akshaynadkarni: Thinking about the event hierarchy a bit more. Even with Option 1, we could separate cause from effect in the event naming:

**Tenant-level (all tiers failed, tenant is Progressing):**
- `TenantNotReady`: emitted when no tier resolves at all

**Tier-level (specific tier failed, tenant may still be Ready):**
- `TenantStorageTierNotReady`: emitted per failing tier, with reason in the message:
  - `DuplicateStorageClass`: multiple SCs matched for the tier
  - `StorageClassMissing`: no SC found for the tier (no tenant-specific, no Default)

This separates what happened (cause) from the impact (effect), and also future-proofs for new failure reasons without changing the event structure. Today `DuplicateStorageClass` serves as both the cause and the event name, which works but conflates the two.

One thing to consider: for `TenantStorageTierNotReady` to be meaningful, we'd need to know which tiers are expected for a tenant. Otherwise we'd fire events for tiers that aren't relevant to the tenant (e.g., a shared Default "archive" tier that the tenant doesn't use). This might circle back to tracking enabled tiers per tenant, either in the spec or derived from discovered SCs. This can be tackled in post v1.0, if we identify the events are noisy or just as part of event hardening. 

WDYT?

akshaynadkarni (approved):
@zszabo-rh Approving this PR. Let's carry forward [this discussion](https://github.com/osac-project/osac-operator/pull/199/changes#r3165174768) offline. 

It will good to get an agreement on the events to generate and add a ticket to the story to address it properly.

PR #204, opened 2026-04-27
MGMT-23976: Add multi-tier storage types to Tenant CRD (additive)

zszabo-rh:
## Summary

- Add `ResolvedStorageClass` struct with `name` and `tier` fields to support multiple storage tiers per tenant (EP [#32](https://github.com/osac-project/enhancement-proposals/pull/32))
- Add `status.storageClasses` list **alongside** the existing `status.storageClass` string — controller populates both
- Add `osac.openshift.io/storage-tier` label constant
- Add `Storage Tiers` printer column (existing `Storage Class` column retained)
- Use `listType=map` with `listMapKey=tier` to enforce one entry per tier

This is a purely additive change — no fields removed, no existing behavior changed. The deprecated `status.storageClass` field will be removed by [MGMT-24139](https://redhat.atlassian.net/browse/MGMT-24139) once all downstream consumers migrate to `status.storageClasses`.

## Test plan

- [x] `make test` passes — existing tests retained, new assertions verify `storageClasses` mirrors `storageClass`
- [x] `make manifests` regenerates CRD with correct schema
- [x] Verified on hypershift1: Tenant reaches Ready with both fields populated, ComputeInstance flow unaffected

Jira: [MGMT-23976](https://redhat.atlassian.net/browse/MGMT-23976)

## Summary by CodeRabbit

* **New Features**
  * Tenants now expose resolved storage per tier so each tenant can show one resolved StorageClass for each storage tier.
  * kubectl/printer output includes a new "Storage Tiers" column showing resolved tier(s) for each tenant.

* **Documentation**
  * The legacy single StorageClass status field is deprecated and retained for backward compatibility; migrate to the new per-tier StorageClasses status.


internal/controller/tenant_controller.go
```diff
@@ -177,10 +181,20 @@ func (r *TenantReconciler) handleUpdate(ctx context.Context, req reconcile.Reque
...
 }
 
+// storageTierFromLabel reads the osac.openshift.io/storage-tier label from a
+// StorageClass. Returns "default" if the label is absent or empty.
+func storageTierFromLabel(sc *storagev1.StorageClass) string {
+	if tier := sc.GetLabels()[osacStorageTierLabel]; tier != "" {
+		return tier
+	}
+	return "default"
+}
```
akshaynadkarni, line 197: When the `storage-tier` label is absent, this falls back to `"default"`. Two questions:

1. What's the motivation for the fallback vs. treating a missing label as an error?
2. When should this fallback be removed?

akshaynadkarni, line 197: [Kubernetes label values](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set) allow uppercase, but the [`Tier` field validation](https://github.com/osac-project/osac-operator/blob/1d91dcab7593ba4404fd6a816700d78eb7724b64/api/v1alpha1/tenant_types.go#L72) only allows lowercase. If someone labels a StorageClass with `storage-tier: FAST`, the controller reads it as-is, finds the StorageClass, and tries to write the status. The API server rejects the update because `FAST` does not match the `Tier` validation pattern. Since the entire status update fails, the conditions do not get written either. The tenant is stuck in Progressing with no indication of what went wrong.

A couple of options:
1. Lowercase the value in `storageTierFromLabel` to handle this silently
2. Validate the value and set a `StorageClassReady=False` condition with a descriptive message before attempting the status update

What do you think?

internal/controller/tenant_controller.go
```diff
@@ -177,10 +181,21 @@ func (r *TenantReconciler) handleUpdate(ctx context.Context, req reconcile.Reque
 // condition that should be set on the Tenant.
 type storageClassResult struct {
 	name    string
+	tier    string
 	reason  string
```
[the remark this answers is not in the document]
zszabo-rh: This additive PR intentionally keeps the old single-tier behavior and wraps the result in a one-element list. The full multi-tier resolution (grouping by tier, per-tier duplicate detection) is part of a separate PR.

internal/controller/tenant_controller.go
zszabo-rh: The fallback exists for backward compatibility during the transition, so that existing tenants on the dev env wouldn't break. The fallback removal is tracked in the cleanup ticket. Once the full multi-tier resolution lands, SCs without the tier label will be treated as errors.

zszabo-rh: You're right, this can be a problem..
Applied strings.ToLower() to prevent the scenario you described!

akshaynadkarni (approved):
Changes LGTM

PR #209, opened 2026-04-28
MGMT-24003: requeue after implementation-strategy annotation update

eranco74:
## Summary
- Fix duplicate AAP job launches caused by a race condition in networking controllers (VirtualNetwork, Subnet, SecurityGroup)
- When the implementation-strategy annotation is set, the metadata `r.Update()` increments resourceVersion, causing subsequent status flushes to fail with a conflict error — the job ID is lost, and the next reconcile triggers a duplicate
- Fix: return `Requeue: true` after the annotation update instead of continuing to provisioning in the same reconcile

## Test plan
- [x] New test: `should requeue after setting implementation-strategy annotation without triggering a job` — verifies first reconcile sets annotation and requeues without calling the provisioning provider, and second reconcile triggers the job
- [x] Updated existing tests that assumed provisioning happens in the same reconcile as annotation update
- [x] Full controller test suite passes (289 specs)
- [x] `go vet` and `gofmt` clean

🤖 Generated with [Claude Code](https://claude.ai/code)

## Summary by CodeRabbit

* **Bug Fixes & Improvements**
  * Controllers for SecurityGroup, Subnet, and VirtualNetwork now update implementation-strategy annotations and return immediately so follow-up processing occurs in the next reconcile cycle, reducing transient errors and improving stability.

* **Tests**
  * Controller tests updated to assert no immediate requeue after annotation writes and to validate subsequent reconciliation triggers expected provisioning and status transitions.

### DIFF
--- a/internal/controller/securitygroup_controller.go
+++ b/internal/controller/securitygroup_controller.go
@@ -177,9 +177,7 @@ func (r *SecurityGroupReconciler) handleUpdate(ctx context.Context, sg *v1alpha1
 		if err := r.Update(ctx, sg); err != nil {
 			return ctrl.Result{}, err
 		}
-		if err := r.Get(ctx, client.ObjectKeyFromObject(sg), sg); err != nil {
-			return ctrl.Result{}, err
-		}
+		return ctrl.Result{}, nil
 	}
 
 	// Compute desired config version from spec and inherited implementation strategy
--- a/internal/controller/securitygroup_controller_test.go
+++ b/internal/controller/securitygroup_controller_test.go
@@ -407,11 +407,18 @@ var _ = Describe("SecurityGroupReconciler", func() {
 				}, nil
 			}
 
-			// First reconcile adds finalizer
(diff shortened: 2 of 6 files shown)


internal/controller/securitygroup_controller.go
```diff
@@ -177,9 +177,7 @@ func (r *SecurityGroupReconciler) handleUpdate(ctx context.Context, sg *v1alpha1
 		if err := r.Update(ctx, sg); err != nil {
 			return ctrl.Result{}, err
 		}
-		if err := r.Get(ctx, client.ObjectKeyFromObject(sg), sg); err != nil {
-			return ctrl.Result{}, err
-		}
+		return ctrl.Result{Requeue: true}, nil
```
adriengentil:
```suggestion
		return ctrl.Result{}, nil
```
should be enough because the `Update` above will already trigger a reconciliation loop on the CR (and `Requeue` is deprecated if I remember well, we should use `RequeueAfter` instead)

adriengentil (approved):

PR #210, opened 2026-04-28
MGMT-23828: Add automatic storage provisioning to Tenant controller

zszabo-rh:
## Summary

- Tenant controller triggers AAP to automatically provision/deprovision tenant StorageClasses
- Backward compatible: pre-existing manually-labeled SCs skip provisioning
- AAP template names: `osac-create-org` / `osac-delete-org`
- Companion PR: [osac-aap#266](https://github.com/osac-project/osac-aap/pull/266) (Ansible role + playbooks + config-as-code)

## Changes

**Tenant controller:**
- `handleStorageProvisioning`: triggers AAP when no labeled SC found, polls job status via `pollProvisionJob`
- `handleStorageDeprovisioning`: triggers AAP on delete, waits for completion before finalizer removal
- StorageClass watch: re-reconciles Tenant when labeled SCs change (self-healing on SC deletion)
- Failed provision jobs do not auto-retry (avoids infinite loops — wait for external trigger)
- Successful provision jobs are treated as stale when SC is missing (enables re-provision after SC deletion)
- `BlockDeletionOnFailure` honored across operator restarts for crash recovery
- Without AAP provider (EDA mode): controller waits for manually-created StorageClass
- Nil-provider guard prevents panic if credentials are removed between restarts

**Provisioning package:**
- Added Tenant to `GetJobsFromResource()` type switch in `provision_lifecycle.go`
- Non-ComputeInstance resources treated as terminal in EDA job path (graceful fallback)

**main.go:**
- `setupTenantController` creates tenant-specific AAP provider with explicit template names
- Rejects unknown `OSAC_PROVISIONING_PROVIDER` values (fail-fast instead of silent nil provider)
- Logs explicitly when EDA provider is configured (tenant provisioning not supported)

**CRD:** Added `status.jobs` to Tenant (additive, no migration needed)

## Deferred

- **Multi-cluster client**: `tenantStorageClassExists` uses local client instead of `targetClient`. All current deployments use `LocalCluster` where both are identical. Will fix when multi-cluster Tenant is tested.
- **Dedicated tenant SA**: Currently uses shared `osac-sa`. Scoped RBAC is an osac-installer change.

## Test plan

- [x] Unit tests: backward compat (multi-tier SC resolution with conditions), provisioning (trigger + poll + Ready), no-provider (requeue), trigger error (failed job, no retry), job failure, delete with provider (deprovisioning), delete without provider (blocks until manual SC removal)
- [x] Existing ComputeInstance/networking tests pass (behavior unchanged)
- [x] Lint: 0 issues
- [x] E2E on hypershift1: create Tenant → AAP provisions SC → Ready → delete SC (self-heal) → delete Tenant → AAP deprovisions → clean deletion

Assisted-By: Claude Code (Anthropic)


internal/controller/tenant_controller.go
```diff
@@ -172,6 +215,240 @@ func (r *TenantReconciler) handleUpdate(ctx context.Context, req reconcile.Reque
...
+// tenantStorageClassExists checks whether at least one StorageClass labeled for
+// this tenant already exists on the cluster.
+func (r *TenantReconciler) tenantStorageClassExists(ctx context.Context, instance *v1alpha1.Tenant) (bool, error) {
+	var scList storagev1.StorageClassList
+	if err := r.List(ctx, &scList, client.MatchingLabels{
+		osacTenantAnnotation: instance.GetName(),
+	}); err != nil {
+		return false, err
+	}
+	return len(scList.Items) > 0, nil
```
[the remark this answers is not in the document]
zszabo-rh: Valid for multi-cluster deployments with OSAC_REMOTE_CLUSTER_KUBECONFIG. All current deployments use  LocalCluster where both clients are identical. Will address as a follow-up when multi-cluster Tenant support is tested end-to-end.

cmd/main.go
```diff
@@ -416,13 +420,49 @@ func setupComputeInstanceControllers(
...
+				"", aapInsecureSkipVerify,
+			)
+			if err != nil {
+				return fmt.Errorf("tenant provisioning provider: %w", err)
+			}
+			setupLog.Info("tenant storage provisioning configured",
+				"provisionTemplate", tenantProvisionTemplate,
+				"deprovisionTemplate", tenantDeprovisionTemplate)
+		}
+	}
```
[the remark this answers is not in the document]
zszabo-rh: Fixed: setupTenantController now uses an explicit switch that rejects unknown provider types with a clear error, and logs an info message for EDA (which doesn't support tenant provisioning).

tzvatot (commented):
## PR #210: MGMT-23828 — Add automatic storage provisioning to Tenant controller

Adds AAP-driven StorageClass provisioning/deprovisioning to the Tenant controller, following the same pattern as ComputeInstance provisioning. The design is well thought out — backward compat with pre-labeled SCs, no-retry on failure (avoiding infinite loops), self-healing via SC watch, and BlockDeletionOnFailure honored across restarts. Test coverage is solid with 7 test cases covering the key paths.

| Category | Count |
|----------|-------|
| 🔴 Critical | 1 |
| 🟡 Important | 2 |
| 💡 Suggestion | 2 |

---

### 🔴 `tenantStorageClassExists` uses local client instead of target cluster client

**File:** `internal/controller/tenant_controller.go` — `tenantStorageClassExists()`

```go
func (r *TenantReconciler) tenantStorageClassExists(ctx context.Context, instance *v1alpha1.Tenant) (bool, error) {
    var scList storagev1.StorageClassList
    if err := r.List(ctx, &scList, client.MatchingLabels{...}); err != nil {
```

`r.List` uses the local management cluster client, but StorageClasses are on the target cluster. The PR description acknowledges this: "Will fix when multi-cluster Tenant is tested." In `LocalCluster` mode both clients are the same, so this works today. But it's a latent bug — when multi-cluster is enabled, `handleDelete` will check the wrong cluster for SCs and either skip deprovisioning (SC exists on target but not local) or spuriously trigger it (SC exists on local but not target).

**Recommendation:** Use the target cluster client now if feasible. If deferred, file a Jira ticket to track the fix — TODO comments tend to get lost.

---

### 🟡 `handleStorageDeprovisioning` doesn't record a failed job when trigger errors

**File:** `internal/controller/tenant_controller.go:535-545`

```go
if err != nil {
    var rateLimitErr *provisioning.RateLimitError
    if errors.As(err, &rateLimitErr) {
        ...
    }
    log.Error(err, "failed to trigger storage deprovisioning")
    return ctrl.Result{RequeueAfter: r.StatusPollInterval}, nil
}
```

When `TriggerDeprovision` fails (non-rate-limit), the error is logged but no failed job is recorded in `status.jobs`. Compare with `handleStorageProvisioning` which records a failed job with empty JobID on trigger failure. This means a failed deprovision trigger is invisible in the CR status — the operator just silently retries on next reconcile with no audit trail.

**Recommendation:** Record a failed deprovision job the same way provision does:

```go
newJob := v1alpha1.JobStatus{
    Type:      v1alpha1.JobTypeDeprovision,
    Timestamp: metav1.NewTime(time.Now().UTC()),
    State:     v1alpha1.JobStateFailed,
    Message:   fmt.Sprintf("Failed to trigger storage deprovisioning: %v", err),
}
instance.Status.Jobs = provisioning.AppendJob(instance.Status.Jobs, newJob, r.MaxJobHistory)
```

---

### 🟡 Race between finalizer add and status update

**File:** `internal/controller/tenant_controller.go:146-151`

```go
if !controllerutil.ContainsFinalizer(instance, tenantFinalizer) {
    controllerutil.AddFinalizer(instance, tenantFinalizer)
    if err := r.Update(ctx, instance); err != nil {
        return ctrl.Result{}, err
    }
}
```

The finalizer is added via `r.Update()` on the metadata, but later in the same reconcile the status is updated via `r.Status().Update()`. If the finalizer update succeeds but the status update hits a conflict (stale resourceVersion), the next reconcile re-reads the object with the finalizer already present but status unchanged — which is fine. However, the finalizer update uses `r.Update()` which also writes status, potentially overwriting status changes from concurrent reconciles. The existing pattern in other controllers uses a separate status update path.

This is the same pattern used by other controllers in this repo, so it's consistent. But worth noting if you ever see status conflicts.

---

### 💡 `aap_provider.go` — EDA fallback for unknown resource types silently succeeds

**File:** `pkg/provisioning/aap_provider.go:228-231`

```go
phase, err := getResourcePhase(resource)
if err != nil {
    log.Info("EDA provision job on unsupported resource type, treating as terminal", "jobID", latestProvisionJob.JobID)
    return true, nil, nil
}
```

When `getResourcePhase` fails (unknown resource type), the code treats the EDA job as terminal and allows deprovisioning. This is pragmatic for now (EDA is only used for CI/ClusterOrder) but could silently allow deprovisioning before provisioning is actually complete if a new resource type is added without updating `getResourcePhase`. A warning log level would be more appropriate than info.

---

### 💡 Test boilerplate — BeforeEach/AfterEach repeated across 5 contexts

**File:** `internal/controller/tenant_controller_test.go`

The namespace creation, tenant creation, and cleanup blocks are nearly identical across all 5 test contexts. A shared helper like `setupTenantTest(ctx, name, withSC bool) (nn, cleanup)` would reduce the 100+ lines of boilerplate.

---

**Overall:** Solid implementation with good test coverage and well-documented deferred items. The multi-cluster client issue is the main concern — it's a known limitation but should at least be commented at the code level. The missing failed-job recording on deprovision trigger is worth fixing for observability parity with the provision path.

internal/controller/tenant_controller.go
```diff
@@ -191,6 +240,260 @@ type tierResolutionResult struct {
...
+			return ctrl.Result{RequeueAfter: r.StatusPollInterval}, nil
+
+		case provisioning.DeprovisionSkipped:
+			log.Info("provider skipped storage deprovisioning")
+			return ctrl.Result{}, nil
+
+		case provisioning.DeprovisionWaiting:
+			log.Info("storage deprovisioning not ready, requeueing")
+			return ctrl.Result{RequeueAfter: r.StatusPollInterval}, nil
+		}
```
akshaynadkarni, line 450: The switch on `result.Action` has no `default` case. If `TriggerDeprovision` returns an unexpected Action value, execution falls through to line 450 where `latestJob.JobID` is accessed. Since we entered the trigger block because `latestJob == nil || latestJob.JobID == ""`, `latestJob` could be nil here, causing a nil-pointer dereference panic.

This requires a provider to return an Action value outside the defined enum, which is unlikely with the current AAP provider, but Go does not enforce exhaustive enum matching.

```suggestion
		case provisioning.DeprovisionWaiting:
			log.Info("storage deprovisioning not ready, requeueing")
			return ctrl.Result{RequeueAfter: r.StatusPollInterval}, nil

		default:
			log.Error(nil, "unexpected deprovision action, requeueing", "action", result.Action)
			return ctrl.Result{RequeueAfter: r.StatusPollInterval}, nil
		}
```

internal/controller/tenant_controller.go
```diff
@@ -191,6 +240,260 @@ type tierResolutionResult struct {
...
+	provisioning.UpdateJob(instance.Status.Jobs, updatedJob)
+
+	if !status.State.IsTerminal() {
+		log.Info("storage provisioning job still running", "jobID", latestJob.JobID, "state", status.State)
+		return ctrl.Result{RequeueAfter: r.StatusPollInterval}, nil
+	}
+
+	if status.State.IsSuccessful() {
+		log.Info("storage provisioning job succeeded, requeueing to confirm StorageClass", "jobID", latestJob.JobID)
+		return ctrl.Result{Requeue: true}, nil
```
akshaynadkarni: When `pollProvisionJob` sees the job succeeded, it returns `Requeue: true` (immediate requeue). If the informer cache hasn't synced the new StorageClass yet, the next reconcile sees no SC, enters `handleStorageProvisioning`, and `needsProvisionJob` returns true for the old Succeeded job, triggering a duplicate provision.

Would a short `RequeueAfter` (e.g., 5s) instead of `Requeue: true` make sense here? That gives the informer time to sync before the next reconcile. Or is the `osac-create-org` playbook idempotent?

config/crd/bases/osac.openshift.io_tenants.yaml
```diff
@@ -163,6 +163,8 @@ spec:
                       enum:
                       - provision
                       - deprovision
+                      - attach
+                      - detach
```
akshaynadkarni, line 167: Is this in `main`? 

I couldn't find it: 
https://github.com/osac-project/osac-operator/blob/main/config/crd/bases/osac.openshift.io_tenants.yaml

I want to make sure you are rebasing with the latest version. Or perhaps I need more coffee ☕ :)

akshaynadkarni: Please disregard this :point_up: comment. I got confused when I saw this cause I am in the process of removing these actions as part of PublicIPAttachment development and am instead using the provision and deprovision ones. 

In short - I needed more coffee :coffee: :)

akshaynadkarni (approved):
Code changes LGTM.

PR #211, opened 2026-04-28
MGMT-23735: Add parent publicippool name annotation on publicip CR

DakCrowder:
Aims to support [MGMT-23735](https://redhat.atlassian.net/browse/MGMT-23907).

The `metallb.universe.tf/address-pool` we need to set in the AAP role needs access to the parent public ip pool k8s resource name.  Currently we only have the fulfillment-service generated uuid in the spec.pool, this adds an additional annotation so we can easily set the metallb value without additional round trips looking that info up.

## Summary by CodeRabbit

* **New Features**
  * PublicIP resources now include an annotation tracking their parent PublicIPPool reference.

* **Improvements**
  * Enhanced annotation update efficiency by consolidating multiple metadata changes into a single operation.

### DIFF
--- a/internal/controller/constants_common.go
+++ b/internal/controller/constants_common.go
@@ -34,6 +34,10 @@ const (
 	// (e.g., "cudn" -> cudn_virtual_network or cudn_subnet role).
 	osacImplementationStrategyAnnotation = osacPrefix + "/implementation-strategy"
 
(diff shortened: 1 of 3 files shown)


akshaynadkarni (approved):
Changes LGTM :)

PR #216, opened 2026-05-04
MGMT-23949: add DeletionTimestamp guard to management-state check

zszabo-rh:
## Summary

- Add `DeletionTimestamp.IsZero()` guard to the management-state unmanaged check in ComputeInstance, ClusterOrder, and Subnet controllers — without this, CRs annotated as `unmanaged` get stuck in `Terminating` forever because `handleDelete` is never called and the finalizer is never removed
- Add the management-state unmanaged check (with the guard) to SecurityGroup and VirtualNetwork controllers, which were missing it entirely, for consistency across all controllers
- Add test cases for all 5 controllers covering both the "ignore when unmanaged" and "still delete when unmanaged" scenarios

Fixes the pattern already corrected in PublicIPPool (osac-project/osac-operator#176) and PublicIP.

**Jira:** [MGMT-23949](https://issues.redhat.com/browse/MGMT-23949)

## Test plan

- [x] `make lint` — 0 issues
- [x] `make test` — 320 specs, all passing
- [x] 12 management-state tests cover all 5 affected controllers (ComputeInstance, ClusterOrder, Subnet, SecurityGroup, VirtualNetwork)
- [ ] Manual verification: annotate a provisioned resource as `unmanaged`, then delete it — should not get stuck in `Terminating`

## Summary by CodeRabbit

* **Bug Fixes**
  * Controllers now still process unmanaged resources when they're being deleted so finalizers run and resources are reliably cleaned up instead of being permanently skipped.

* **Tests**
  * Added tests covering deletion and reconcile behavior for unmanaged resources across controllers, validating finalizer handling and eventual resource removal.

[Review Change Stack](https://app.coderabbit.ai/change-stack/osac-project/osac-operator/pull/216)


akshaynadkarni (commented):
~~This comment was posted at the wrong level. See the inline comment on clusterorder_controller_test.go#L435-L439 instead.~~

internal/controller/clusterorder_controller_test.go
```diff
@@ -395,4 +395,49 @@ var _ = Describe("ClusterOrder Controller", func() {
...
+			_, err := controllerReconciler.Reconcile(ctx, reconcile.Request{
+				NamespacedName: key,
+			})
+			Expect(err).NotTo(HaveOccurred())
+
+			Eventually(func() bool {
+				return client.IgnoreNotFound(
+					k8sClient.Get(ctx, key, &v1alpha1.ClusterOrder{}),
+				) == nil
+			}, 5*time.Second, 100*time.Millisecond).Should(BeTrue())
```
akshaynadkarni, line 437: The deletion assertion `client.IgnoreNotFound(Get(...)) == nil` is always true and does not verify that the object was actually garbage collected.

`client.IgnoreNotFound` takes an error and returns it as-is, unless it is a NotFound error, in which case it returns nil. So when used with `Get`:

- **Object still exists:** `Get` succeeds, returns `nil`. `IgnoreNotFound(nil)` returns `nil`. `nil == nil` is `true`.
- **Object was deleted:** `Get` fails with NotFound. `IgnoreNotFound(NotFound)` returns `nil`. `nil == nil` is `true`.
- **Some other error:** `Get` returns `otherError`. `IgnoreNotFound(otherError)` returns `otherError`. `otherError == nil` is `false`.

The first two cases both produce `nil`, so the assertion passes whether the object exists or not. The test can only fail on an unexpected error (e.g., network issue), which doesn't happen in envtest. This means the test would pass even if the finalizer was never removed and the CR was stuck in Terminating.

The correct pattern is `errors.IsNotFound(k8sClient.Get(...))`, which returns `true` **only** when Get returns NotFound (the object is actually gone). If the object still exists, Get returns nil, and `IsNotFound(nil)` is `false`, so the Eventually times out and the test fails, which is what we want.

The ComputeInstance test in this same PR already uses the correct pattern:
https://github.com/osac-project/osac-operator/blob/fb799f708b1588e143ad05ad06a5f139547577fa/internal/controller/computeinstance_controller_test.go#L222-L224

```go
Eventually(func() bool {
    return errors.IsNotFound(k8sClient.Get(ctx, key, &osacv1alpha1.ComputeInstance{}))
}, 5*time.Second, 100*time.Millisecond).Should(BeTrue())
```

Please apply the same pattern to these three locations:
- [clusterorder_controller_test.go#L435-L439](https://github.com/osac-project/osac-operator/blob/fb799f708b1588e143ad05ad06a5f139547577fa/internal/controller/clusterorder_controller_test.go#L435-L439) (this file)
- [subnet_controller_test.go#L290-L294](https://github.com/osac-project/osac-operator/blob/fb799f708b1588e143ad05ad06a5f139547577fa/internal/controller/subnet_controller_test.go#L290-L294)
- [virtualnetwork_controller_test.go#L523-L527](https://github.com/osac-project/osac-operator/blob/fb799f708b1588e143ad05ad06a5f139547577fa/internal/controller/virtualnetwork_controller_test.go#L523-L527)

Note: the `errors` import (`apierrors "k8s.io/apimachinery/pkg/api/errors"`) is already present in the ComputeInstance test file but may need to be added to the ClusterOrder, Subnet, and VirtualNetwork test files.

internal/controller/securitygroup_controller_test.go
```diff
@@ -628,6 +628,73 @@ var _ = Describe("SecurityGroupReconciler", func() {
...
+			Expect(fakeClient.Get(ctx, key, fetched)).To(Succeed())
+			now := metav1.Now()
+			fetched.DeletionTimestamp = &now
+
+			Expect(fetched.ObjectMeta.DeletionTimestamp.IsZero()).To(BeFalse())
+
+			// The fake client's Update will fail because DeletionTimestamp is
+			// immutable, but the important assertion is that the finalizer was
+			// removed from the in-memory object.
+			_, _ = reconciler.handleDelete(ctx, fetched)
```
akshaynadkarni: This test calls `handleDelete` directly rather than going through `Reconcile`. The fix in this PR adds the [`DeletionTimestamp.IsZero()` guard in `Reconcile()`](https://github.com/osac-project/osac-operator/blob/fb799f708b1588e143ad05ad06a5f139547577fa/internal/controller/securitygroup_controller.go#L106), but this test bypasses that code path entirely.

What this means: the test proves that `handleDelete` removes the finalizer on an unmanaged resource, but it doesn't prove that `Reconcile` would actually reach `handleDelete` when the unmanaged annotation is set. If someone accidentally removed or broke the [`DeletionTimestamp.IsZero()` guard](https://github.com/osac-project/osac-operator/blob/fb799f708b1588e143ad05ad06a5f139547577fa/internal/controller/securitygroup_controller.go#L106) (reverting it back to `if exists && val == ManagementStateUnmanaged`), this test would still pass.

The other four controllers all test through `Reconcile`, which exercises the full flow. For example, the [ComputeInstance test](https://github.com/osac-project/osac-operator/blob/fb799f708b1588e143ad05ad06a5f139547577fa/internal/controller/computeinstance_controller_test.go#L217-L224) calls `Reconcile`, and the [Subnet test](https://github.com/osac-project/osac-operator/blob/fb799f708b1588e143ad05ad06a5f139547577fa/internal/controller/subnet_controller_test.go#L285-L295) does the same.

Is the direct [`handleDelete` call](https://github.com/osac-project/osac-operator/blob/fb799f708b1588e143ad05ad06a5f139547577fa/internal/controller/securitygroup_controller_test.go#L668) intentional here (to stay consistent with how the existing SecurityGroup deletion tests use the fake client), or would it be worth switching to a `Reconcile`-level test to match the other controllers?

internal/controller/clusterorder_controller_test.go
zszabo-rh: Switched all three locations to errors.IsNotFound() to match the ComputeInstance pattern!

internal/controller/securitygroup_controller_test.go
zszabo-rh: Switched this test to use envtest (k8sClient) and go through Reconcile..

akshaynadkarni (approved):
Thanks for addressing the comments. 

Changes LGTM.

PR #217, opened 2026-05-04
MGMT-23989: Add publicip target namespace annotation

DakCrowder:
[MGMT-23989](https://redhat.atlassian.net/browse/MGMT-23989)

Sets a `osac.openshift.io/publicip-target-namespace` annotation on the Public IP CR when `spec.computeInstance` is set.  This annotation will be used by AAP roles to know where to create the associated LoadBalancer Service for the given Public IP.

## Summary by CodeRabbit

* **New Features**
  * PublicIP resources can now associate with ComputeInstance resources and automatically synchronize the target namespace annotation from the referenced instance.
  * Controller now watches ComputeInstance changes to keep PublicIP state up to date.

* **Tests**
  * Added tests validating annotation set/clear/requeue behaviors and mapping of ComputeInstances to affected PublicIPs.

### DIFF
--- a/internal/controller/publicip_controller.go
+++ b/internal/controller/publicip_controller.go
@@ -51,20 +51,22 @@ const (
 // Phase transitions: "" -> Progressing -> Ready/Failed; on delete: Deleting.
 type PublicIPReconciler struct {
 	client.Client
-	APIReader            client.Reader
-	Scheme               *runtime.Scheme
-	mgr                  mcmanager.Manager
-	NetworkingNamespace  string
-	ProvisioningProvider provisioning.ProvisioningProvider
-	StatusPollInterval   time.Duration
-	MaxJobHistory        int
-	targetCluster        mc.ClusterName
+	APIReader                client.Reader
+	Scheme                   *runtime.Scheme
+	mgr                      mcmanager.Manager
+	NetworkingNamespace      string
+	ComputeInstanceNamespace string
+	ProvisioningProvider     provisioning.ProvisioningProvider
+	StatusPollInterval       time.Duration
+	MaxJobHistory            int
+	targetCluster            mc.ClusterName
 }
 
 // NewPublicIPReconciler creates a new reconciler for PublicIP resources.
 func NewPublicIPReconciler(
 	mgr mcmanager.Manager,
 	networkingNamespace string,
+	computeInstanceNamespace string,
 	provisioningProvider provisioning.ProvisioningProvider,
 	statusPollInterval time.Duration,
 	maxJobHistory int,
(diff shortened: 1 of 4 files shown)


internal/controller/publicip_controller.go
```diff
@@ -240,6 +256,55 @@ func (r *PublicIPReconciler) handleUpdate(ctx context.Context, publicIP *v1alpha
...
+	}
+
+	ci := &ciList.Items[0]
+	if ci.Status.TenantReference == nil {
+		log.Info("ComputeInstance has no TenantReference yet, requeueing",
+			"computeInstance", ci.Name, "computeInstanceUUID", publicIP.Spec.ComputeInstance)
+		return false, true, nil
+	}
+
+	targetNamespace := ci.Status.TenantReference.Namespace
```
akshaynadkarni: Hey, I think the ticket captured this incorrectly, and some things have changed since it was written. Sorry about that.

`ci.Status.TenantReference.Namespace` is the namespace where the Tenant CR lives (e.g., `osac-tenant`), not the namespace where the VM runs. It's set as `tenant.GetNamespace()` in [`computeinstance_resources.go:52`](https://github.com/osac-project/osac-operator/blob/3e7dce7f82712ea8f7e4cb3093b34f55bf34989c/internal/controller/computeinstance_resources.go#L52).

The actual namespace where the LoadBalancer Service should be created depends on whether the CI has a subnet or not. `ci.Status.VirtualMachineReference.Namespace` already resolves this correctly in both cases (set at [`computeinstance_controller.go:819`](https://github.com/osac-project/osac-operator/blob/3e7dce7f82712ea8f7e4cb3093b34f55bf34989c/internal/controller/computeinstance_controller.go#L819) from the VM's actual namespace, which accounts for subnet vs. tenant at [`L661-663`](https://github.com/osac-project/osac-operator/blob/3e7dce7f82712ea8f7e4cb3093b34f55bf34989c/internal/controller/computeinstance_controller.go#L661-L663)).

This would also match the resolution AAP does with `compute_instance_target_namespace` (see [`ocp_virt_vm/tasks/create.yaml`](https://github.com/osac-project/osac-aap/blob/main/collections/ansible_collections/osac/templates/roles/ocp_virt_vm/tasks/create.yaml#L14)).

akshaynadkarni (changes requested):
Please see my comment [here](https://github.com/osac-project/osac-operator/pull/217/changes#r3183911921)

Still reviewing the rest of the PR.

internal/controller/publicip_controller.go
akshaynadkarni: Also, can you add a `Watches()` on ComputeInstance in `SetupWithManager`, similar to how the ComputeInstance controller watches Tenant and VirtualMachine resources ([`computeinstance_controller.go:232-248`](https://github.com/osac-project/osac-operator/blob/3e7dce7f82712ea8f7e4cb3093b34f55bf34989c/internal/controller/computeinstance_controller.go#L232-L248))?

Today a PublicIP is attached to a CI that already has a running VM, so `VirtualMachineReference.Namespace` would already be populated. But in the future when a ComputeInstance automatically gets a PublicIP assigned at creation time, the VM won't exist yet and the controller would be polling every 10 seconds until it comes up. A Watch would let the PublicIP reconcile immediately when the CI's status changes.

akshaynadkarni (approved):
Thanks for addressing the comments. Changes LGTM.

PR #222, opened 2026-05-05
MGMT-24300: fix duplicate AAP jobs from feedback controller race condition

danmanor:
The feedback controller adds its finalizer concurrently with the resource controller's reconcile, changing the resourceVersion. When the resource controller tries to flush status after triggering an AAP job, it gets a conflict error and the job is never recorded. The next reconcile sees no job and triggers a duplicate.

Fix by re-reading the resource from the API server via APIReader.Get() before status updates to get the latest resourceVersion. Applied to all 4 networking controllers (VirtualNetwork, Subnet, SecurityGroup, PublicIPPool) in both the statusFlush callback and the end-of-reconcile status update. Added concurrent-modification tests for all controllers.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced stability of status updates across cloud resource controllers by implementing conflict-aware retry mechanisms. Status changes now persist reliably even when resources are concurrently modified, preventing loss of provisioning job information.

* **Tests**
  * Added test coverage verifying status persistence for PublicIPPool, SecurityGroup, Subnet, and VirtualNetwork resources during concurrent resource modifications.

### DIFF
--- a/internal/controller/publicippool_controller.go
+++ b/internal/controller/publicippool_controller.go
@@ -24,6 +24,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/util/retry"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	controllerutil "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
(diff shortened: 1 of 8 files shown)


internal/controller/publicippool_controller.go
```diff
@@ -129,6 +129,12 @@ func (r *PublicIPPoolReconciler) Reconcile(ctx context.Context, req mcreconcile.
 
 	if !equality.Semantic.DeepEqual(pool.Status, *oldstatus) {
 		log.Info("status requires update")
+		fresh := &v1alpha1.PublicIPPool{}
```
trewest: It looks like there is an example of how to do this in the codebase already - https://github.com/osac-project/osac-operator/blob/main/internal/controller/computeinstance_controller.go#L163-177. If you add those methods they can be used like:
```
  // End-of-reconcile:
  if !equality.Semantic.DeepEqual(pool.Status, *oldstatus) {
      log.Info("status requires update")
      if err := r.updateStatusWithRetry(ctx, req.NamespacedName, pool.Status); err != nil {
          return res, err
      }
  }

  // statusFlush callback in handleProvisioning:
  func() error {
      return r.updateStatusWithRetry(ctx, client.ObjectKeyFromObject(pool), pool.Status)
  },
```

danmanor: nice, will do it like that, thanks

trewest (approved):

PR #223, opened 2026-05-05
Relocate AAP Client Package

### LINKED ISSUE #387: Move osac-operator's AAP client to a central location
osac-operator's AAP client package is located in `osac-operator/internal/aap`. Go doesn't allow other packages importing it, which is needed by the `bare-metal-fulfillment-operator` and `host-management-openstack` repos to run templates that configure `BareMetalPools` and `HostLeases`.

ajamias:
The bare-metal-fulfillment-operator wants to reuse the code in this package for its way of setting up/tearing down clusters.

Closes osac-project/issues#387

## Summary by CodeRabbit

* **Refactor**
  * Moved core provisioning and AAP code into a public package layout, updating internal references to the new package paths.

* **Chores**
  * Updated lint configuration to exempt certain rules for the new package path.
  * Adjusted documentation to reflect the reorganized file layout.

* **Tests**
  * Updated test imports and mocks to align with the new package layout; test logic unchanged.

[Review Change Stack](https://app.coderabbit.ai/change-stack/osac-project/osac-operator/pull/223)


adriengentil:
I would put it in a `pkg` subfolder as recommended in https://github.com/golang-standards/project-layout#pkg

larsks (changes requested):
Don't you also need to relocate the ProvisioningProvider interface (and in fact, probably just all of `internal/provisioning`)? Any functions or variables that receive an AAPClient interface should in fact be of type ProvisioningProvider (so that the AAP client can be swapped out for some other provisioning backend).

ajamias:
You're right, I previously mistaken the purpose of the ProvisioningProvider interface as solely specific to CaaS without BMaaS, but now realize that it can be used for the bare metal template case as well.

larsks (approved):

PR #227, opened 2026-05-06
OSAC-506: Use separate public / internal ip address fields when syncing CIs

DakCrowder:
In conjunction with https://github.com/osac-project/osac-aap/pull/290, this PR seeks to no longer use the `floating-ip-address` annotation and instead uses a new `public-ip-address` annotation.  Additionally the compute instance status now has explicit `internal_ip_address` and `external_ip_address` fields, and the sync is updated accordingly.

One other small note on behavior is AAP manages the `public-ip-address` annotation and deletes it when a public ip is detached from a compute instance.  In this case we write the empty value back to the `fulfillment-service` to ensure that state is properly reflected.

## Release Notes

* **Bug Fixes**
  * Improved compute instance IP address synchronization to properly update both public and internal IP addresses from their respective sources.

* **Tests**
  * Added comprehensive test coverage for IP address synchronization scenarios.


SiddarthR56 (approved):

akshaynadkarni:
@DakCrowder Could you create a follow-up ticket for adding a `status.publicIPAddress` field to the ComputeInstance CR? Perhaps it can be tackled next week.

Currently the public IP is the only state data the feedback controller reads from `.Annotations` rather than `.Status`. Every other field it syncs to the fulfillment service (phase, conditions, internal IP, lastRestartedAt) comes from the CR's status.

If the resource controller copied the annotation value (or looked up the attached PublicIP CR's `status.address` directly) into `status.publicIPAddress`, the feedback controller would read from status like it does for everything else. This would:

- Make the public IP visible via `kubectl get ci` with a printcolumn
- Keep the feedback controller consistent: it only reads `.Status` for state data
- Open the door to removing the annotation component from osac-aap entirely, since the operator could source the public IP from the PublicIP CR's `status.address` instead

akshaynadkarni (approved):
Changes LGTM. 

Please see [my comment](https://github.com/osac-project/osac-operator/pull/227#issuecomment-4393847813) for some follow-up work. This can be tackled next week.

DakCrowder:
@akshaynadkarni ack'd, will create that ticket!

PR #231, opened 2026-05-08
MGMT-24034: Enable VM access on non-dedicated cluster mode

sk-ilya:
## Summary by CodeRabbit

* **New Features**
  * Console access to virtual machine instances via the management console is now enabled for interactive troubleshooting and administration.

* **Bug Fixes**
  * More accurate detection of whether a desired configuration is applied, preventing false positives after transient spec changes.
  * Improved provisioning flow and status updates so restart and provisioning states are reported and handled more reliably.

[Review Change Stack](https://app.coderabbit.ai/change-stack/osac-project/osac-operator/pull/231)


adriengentil (approved):

eranco74 (commented):
LGTM — the A-B-A config version fix is clean and well-tested. Switching `IsConfigApplied` to only check the latest job per type is the right approach, and the always-delegate pattern in the controller avoids the short-circuit that caused the false positive.

One minor note: in the `else` branch (config not applied), if `RestartRequestedAt` is nil, `RestartInProgress` is never explicitly set to `False`. If a previous reconcile set it to `True` and then config became "not applied" for an unrelated reason, the condition could linger. Minor edge case — worth a follow-up but not a blocker.

PR #238, opened 2026-05-13
Make subnet immutable and security groups mutable in networkAttachments

ori-amizur:
This change enforces immutability on subnet references while allowing security groups to be modified in ComputeInstance networkAttachments.

Changes:
- Add field-level CEL validation on NetworkAttachment.subnetRef (self == oldSelf)
- Change array-level validation to prevent add/remove (size check only)
- Add +listType=map and +listMapKey=subnetRef to make array correlatable
- Add maxItems=8 to reduce CEL cost and stay within validation budget
- Add tests documenting the validation behavior
- Update sample YAML with networkAttachments example

Legacy fields (spec.subnet, spec.securityGroups) remain unchanged for backward compatibility.

Validation enforcement:
- Subnet cannot be changed after creation (structural/placement field)
- Security groups can be modified (policy field)
- Network attachments cannot be added or removed

## Summary by CodeRabbit

* **New Features**
  * Network attachments: subnet references are immutable, security groups remain mutable, attachment count is fixed (max 8) and additions/removals are disallowed.

* **Documentation**
  * Sample manifest and schema docs updated to explain network attachment constraints and multi‑NIC configuration.

* **Tests**
  * Added comprehensive tests validating attachment selection, immutability rules, max‑8 limit, and controller validation behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/osac-project/osac-operator/pull/238?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/config/crd/bases/osac.openshift.io_computeinstances.yaml
+++ b/config/crd/bases/osac.openshift.io_computeinstances.yaml
@@ -134,13 +134,32 @@ spec:
                 description: |-
                   NetworkAttachments defines multiple NICs when more than one subnet (and optional security groups per NIC) is required.
                   When non-empty, subnetRef must be empty; the first entry is the primary subnet for VM placement (subnet-namespace annotation).
+                  Subnet references (per NetworkAttachment) are immutable but security groups can be changed.
+
+                  Why immutability is required:
+                  Currently, ComputeInstances are created with a single NIC. The VM instance must be created in the namespace
+                  of the subnet it uses (the subnet-namespace annotation determines VM placement). This architectural constraint
+                  means that changing the subnet would require recreating the VM in a different namespace, which is not supported.
+                  Similarly, the array size cannot change because adding/removing NICs would require changing the primary subnet
+                  or the VM's namespace, both of which are immutable after VM creation.
+
+                  Immutability enforcement:
+                  - The +listType=map with +listMapKey=subnetRef markers correlate entries by subnetRef for updates
+                  - The size check below prevents adding or removing entries
+                  - The per-item subnetRef validation (self == oldSelf) prevents changing keys of existing entries
+                  Combined, these ensure only securityGroupRefs can be modified.
+
+                  MaxItems is required for CEL cost budget calculation: NetworkAttachment.SubnetRef has a CEL validation
+                  rule (self == oldSelf for immutability). Without maxItems, Kubernetes cannot bound the cost of iterating
+                  over array items to validate this rule, causing CRD installation to fail with "estimated rule cost exceeds budget".
                 items:
                   description: 'NetworkAttachment defines one NIC: a Subnet CR on
                     the hub plus optional SecurityGroup CR names.'
                   properties:
                     securityGroupRefs:
-                      description: SecurityGroupRefs lists SecurityGroup CR names
-                        in the same namespace as the ComputeInstance.
+                      description: |-
+                        SecurityGroupRefs lists SecurityGroup CR names in the same namespace as the ComputeInstance.
+                        Mutable - can be changed to add/remove security groups without recreating the VM.
                       items:
                         type: string
                       type: array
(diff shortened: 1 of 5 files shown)


config/crd/bases/osac.openshift.io_computeinstances.yaml
```diff
@@ -149,13 +151,20 @@ spec:
...
                   type: object
+                maxItems: 8
                 type: array
+                x-kubernetes-list-map-keys:
+                - subnetRef
+                x-kubernetes-list-type: map
                 x-kubernetes-validations:
-                - message: networkAttachments is immutable
-                  rule: self == oldSelf
+                - message: cannot add or remove network attachments
```
tzvatot, line 183: `size(self) == size(oldSelf)` only checks array length. In practice this is caught by the `+listType=map` + `+listMapKey=subnetRef` correlation. Consider commenting that so future readers will not assume the size check is actually limiting the adding/removal of the network attachments.

tzvatot (changes requested):

tzvatot (commented):
### Code review

Good approach — replacing the coarse `self == oldSelf` on the whole array with granular field-level immutability via `+listType=map` and per-field CEL rules. The design correctly separates structural fields (subnet — immutable) from policy fields (security groups — mutable). CRD changes and sample YAML are clean.

| Category | Count |
|----------|-------|
| 🔴 Critical | 0 |
| 🟡 Important | 2 |
| 💡 Suggestion | 1 |

---

🟡 **`+listMapKey=subnetRef` creates a uniqueness constraint that may not be intended**

`api/v1alpha1/computeinstance_types.go:155`

Using `+listType=map` with `+listMapKey=subnetRef` makes `subnetRef` the map key, which means the API server will reject two NetworkAttachments with the same `subnetRef`. If a VM legitimately needs two NICs on the same subnet (e.g., different security group profiles on the same network), this prevents it.

If same-subnet duplicates should be allowed, use `+listType=atomic` with a custom CEL rule for size immutability, or add a second listMapKey (e.g., a user-provided name field) to make the composite key unique.

If same-subnet duplicates are intentionally disallowed, add a comment documenting this constraint — it's a non-obvious side effect of the map key choice.

🟡 **Tests don't exercise the CEL validation — they test Go struct mutability**

`api/v1alpha1/computeinstance_types_test.go`

The "immutability validation" tests construct Go structs and mutate fields in-memory. This proves Go structs are mutable (which is always true) but doesn't validate the CEL rules. The `documents subnet immutability enforcement via CEL` and `documents max network attachments limit` tests are effectively documentation comments wrapped in test syntax — they assert trivially true things (`Expect(8).To(BeNumerically(">", 0))`).

CEL validation can only be tested against a real or envtest API server. If envtest is available in this repo, consider adding an integration test that creates a ComputeInstance and attempts to update the subnetRef — verifying the API server rejects it. If envtest isn't set up, these tests are fine as documentation but should be named accordingly.

💡 **`maxItems=8` — is 8 the right number?**

`api/v1alpha1/computeinstance_types.go:153`

8 is a magic number — is it derived from a product requirement (max NICs per VM), an infrastructure limit (OpenStack port limit), or purely the CEL budget? A comment explaining the rationale would help future maintainers decide whether to change it.

config/crd/bases/osac.openshift.io_computeinstances.yaml
ori-amizur: OK.  Will comment

ori-amizur: Currently we can create a computeinstance with only a single NIC.  Also the VM instance must be created in the subnet namespace.  This dictates that the subnet cannot be changed and also the size cannot change.

ori-amizur: BTW,  without the size check it is still allowed to add/remove network attachments

PR #269, opened 2026-06-01
OSAC-179: remove deprecated status.storageClass field

zszabo-rh:
## Summary

Remove the deprecated `status.storageClass` (singular string) field from the Tenant CRD. All consumers have migrated to `status.storageClasses` (plural list with per-tier resolution).

### Why this is not a breaking change

The `status.storageClass` field was a **status subresource** (observed state set by the controller), not a spec field. No external system writes to it. All consumers that previously read it have been migrated:

- **osac-aap** `tenant_storage_class` role: migrated to `storageClasses` list in commit `77c7956` (May 7) — reads from extra_vars injected by the CI controller, no longer queries the Tenant CR directly
- **computeinstance_controller**: only ever used the plural `StorageClasses` field (via the `StorageClassReady` condition)
- **No external API consumers**: the field was never exposed via the fulfillment-service gRPC/REST API — it only existed on the Kubernetes CR

The singular field was kept temporarily to avoid breaking running dev environments on hypershift1 during the multi-tier migration. That migration is complete.

## Changes

- Remove `StorageClass` field from `TenantStatus` struct and its kubebuilder print column
- Replace the "Storage Class" printer column with "Storage Classes" showing `storageClasses[*].name`
- Remove backward-compatibility sync logic in tenant controller (`handleUpdate`)
- Remove corresponding test assertions
- Regenerate CRD manifests (`config/crd/` and `charts/operator-crds/`)

## Testing

- `go build ./...` — passes
- `go vet ./...` — passes
- `make manifests generate` — passes, CRD YAML verified clean
- Existing unit tests updated (removed assertions for deleted field)
- grep verified: no remaining references to singular `status.storageClass` in any OSAC repo

## Jira

https://redhat.atlassian.net/browse/OSAC-179

---
Generated-by: Claude Code (claude-opus-4-6)

### DIFF
--- a/api/v1alpha1/tenant_types.go
+++ b/api/v1alpha1/tenant_types.go
@@ -81,11 +81,6 @@ type TenantStatus struct {
 	// Namespace is the namespace allocated to the tenant on the target cluster
 	Namespace string `json:"namespace,omitempty"`
 
-	// StorageClass is the StorageClass allocated to the tenant on the target cluster.
-	// Deprecated: use StorageClasses instead. Retained for backward compatibility
-	// until all consumers migrate. Will be removed by MGMT-24139.
-	StorageClass string `json:"storageClass,omitempty"`
-
 	// StorageClasses lists all resolved StorageClass mappings for the tenant,
 	// one per storage tier.
 	// +kubebuilder:validation:Optional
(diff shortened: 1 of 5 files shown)


akshaynadkarni (approved):
Changes LGTM. 

Please look at my question [here](https://github.com/osac-project/osac-operator/pull/269#issuecomment-4598630140). It can be addressed separately.

PR #290, opened 2026-06-10
OSAC-721: add guestOSFamily field to ComputeInstance CRD

ygalblum:
- Add GuestOSFamily string field to ComputeInstanceSpec after Image field
- Field is optional (omitempty) with no default value
- Immutable after creation via CEL validation (self == oldSelf)
- Freeform string (no enum, no pattern) for future OS expansion
- Add OS printcolumn between Memory and RunStrategy
- Regenerate CRD manifests and DeepCopy methods
- Sync Helm CRD templates
- Add 9 validation tests covering:
  - Optional field handling (4 cases): creation without field, linux, windows, freeform string
  - Immutability (3 cases): reject changes, allow updates when unchanged, allow updates when unset
  - Serialization (2 cases): round-trip when set, omitempty when empty
- All existing tests pass (no regressions)

## Summary by CodeRabbit

* **New Features**
  * Added optional `guestOSFamily` field to ComputeInstance resources, allowing users to specify the guest operating system type (Windows or Linux)
  * Guest OS family is immutable after creation and displayed in resource listings as an OS column

* **Tests**
  * Added validation tests for guest OS family field behavior and immutability constraints

### DIFF
--- a/api/v1alpha1/computeinstance_types.go
+++ b/api/v1alpha1/computeinstance_types.go
@@ -100,6 +100,14 @@ type ComputeInstanceSpec struct {
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="image is immutable"
 	Image ImageSpec `json:"image"`
 
+	// GuestOSFamily specifies the guest operating system family for the VM.
+	// To support future OS families, this field is a freeform string.
+	// Currently, any value other than "windows" will be treated as "linux".
+	// This field is immutable after creation.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="guestOSFamily is immutable"
+	GuestOSFamily string `json:"guestOSFamily,omitempty"`
+
 	// Cores is the number of CPU cores
 	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:Minimum=1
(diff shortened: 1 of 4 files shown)


api/v1alpha1/computeinstance_types.go
```diff
@@ -100,6 +100,13 @@ type ComputeInstanceSpec struct {
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="image is immutable"
 	Image ImageSpec `json:"image"`
 
+	// GuestOSFamily specifies the guest operating system family for the VM.
+	// Valid values are "linux" (default) and "windows". Unknown values fall through to linux behavior.
```
vladikr: The text says that there is a default and some valid values but I don't see any `+kubebuilder:default`

ygalblum: If the key is not set or set to an unknown value, the execution path will take the `linux` path. But, it is not intended to set the value to `linux`. So, I'm not sure it's correct to programatically declare it as default. Makes sense?

vladikr: It does! :) I get it, you don't want to mutate the object and you want the runtime to fallback to linux.  
I was simply pointing out that the godoc comment implies that there are valid and default values, which is misleading, from my pov.  
I wonder if, instead, the comment could say that this is a free-form string (to allow future OS families?) and that the operator will use linux behavior when provisioning the VirtualMachine object. 
What do you think?

ygalblum: I've changed is to:
```
// To support future OS families, this field is a freeform string.
// Currently, any value other than "windows" will be treated as "linux".
```

vladikr: Thanks!

charts/operator-crds/templates/osac.openshift.io_computeinstances.yaml
```diff
@@ -106,6 +109,15 @@ spec:
...
                   rule: self == oldSelf
+              guestOSFamily:
+                description: |-
+                  GuestOSFamily specifies the guest operating system family for the VM.
+                  Valid values are "linux" (default) and "windows". Unknown values fall through to linux behavior.
+                  This field is immutable after creation.
+                type: string
+                x-kubernetes-validations:
+                - message: guestOSFamily is immutable
+                  rule: self == oldSelf
```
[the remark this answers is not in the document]
ygalblum: @coderabbitai resolve

omer-vishlitzky (approved):

PR #299, opened 2026-06-17
OSAC-23: Extract storage lifecycle into OSAC Storage Controller

zszabo-rh:
## Summary

Extracts storage provisioning and deprovisioning logic from the Tenant controller into a dedicated OSAC Storage Controller, as specified in the [design document](https://github.com/osac-project/enhancement-proposals/pull/58) (merged) and [PRD](https://github.com/osac-project/enhancement-proposals/pull/52) (merged).

**Key changes:**
- New `StorageReconciler` watches Tenant CRs as a secondary controller with its own finalizer (`osac.openshift.io/storage`)
- Owns two conditions: `StorageBackendReady` (Stage 1: hub Secret) and `ClusterStorageReady` (Stage 2: StorageClasses on target cluster)
- Two independent AAP provider instances: `BackendProvider` and `ClusterStorageProvider` with separate template env vars
- New status types: `StorageBackendStatus` and `ClusterStorageStatus` for future multi-backend/multi-cluster support
- Tenant controller simplified to namespace + UDN management only (zero storage logic)
- New controller enable flag: `OSAC_ENABLE_STORAGE_CONTROLLER` / `--enable-storage-controller`
- AAP client fix: uses numeric template ID when available (resolves launch-by-name issues with non-default AAP projects)

**Env vars:**
```
OSAC_STORAGE_BACKEND_AAP_PROVISION_TEMPLATE    (default: osac-create-tenant-storage-backend)
OSAC_STORAGE_BACKEND_AAP_DEPROVISION_TEMPLATE  (default: osac-delete-tenant-storage-backend)
OSAC_STORAGE_CLUSTER_AAP_PROVISION_TEMPLATE    (default: osac-create-tenant-cluster-storage)
OSAC_STORAGE_CLUSTER_AAP_DEPROVISION_TEMPLATE  (default: osac-delete-tenant-cluster-storage)
```

**Dependencies:**
- **AAP playbook split**: [osac-aap PR #338](https://github.com/osac-project/osac-aap/pull/338) must be merged before enabling the storage controller (playbook filenames must match template names)
- **Merge order**: This PR first (operator), then AAP PR #338. This is safe because the storage controller is **disabled by default** (`OSAC_ENABLE_STORAGE_CONTROLLER=false`) — it won't attempt to launch any templates until explicitly enabled. Merging AAP first would be risky: the config-as-code sync would rename the existing templates, breaking the current tenant controller on any deployment that has it enabled.

## Test plan

- [x] 427/428 unit tests pass (1 pre-existing flaky race condition)
- [x] E2E tested on beaker SNO (edge-17) with VAST appliance — 16 tests across 5 phases (June 12), 6 re-tests after design alignment (June 15)
- [x] Full tenant lifecycle: create → backend provision → cluster storage → delete
- [x] Management-state Unmanaged: skips reconciliation, deletion still runs
- [x] Playbook idempotency verified for all 4 lifecycle actions
- [x] CRD upgrade path: existing tenants auto-reconcile with new condition names

## Summary by CodeRabbit

* **New Features**
  * Introduced Storage Controller for dedicated storage provisioning lifecycle management
  * Added separate storage backend and cluster storage readiness tracking
  * Enhanced status visibility with new printer columns for backend and cluster storage readiness

* **Bug Fixes / Improvements**
  * Standardized job history field naming across resources for consistency
  * Refined Tenant Controller to focus on namespace readiness
  * Improved storage status reporting with dedicated backend and cluster storage status arrays


internal/controller/tenant_controller.go
```diff
@@ -45,22 +39,20 @@ import (
...
 	"github.com/osac-project/osac-operator/api/v1alpha1"
-	"github.com/osac-project/osac-operator/pkg/provisioning"
 )
 
 const tenantFinalizer = "osac.openshift.io/tenant"
 
-// TenantReconciler reconciles a Tenant object
+// TenantReconciler reconciles a Tenant object.
+// Manages namespace creation, UDN reconciliation, and tenant lifecycle.
+// Storage provisioning is handled by the OSAC Storage Controller.
```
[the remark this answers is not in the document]
zszabo-rh: The upstream tenant controller has the same pattern — it only `Get`s the namespace, it doesn't create it. Namespace creation is handled externally (by the fulfillment-service or admin). The refactored tenant controller preserves the exact same behavior.

config/rbac/role.yaml
```diff
@@ -17,6 +17,14 @@ rules:
...
   - update
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - get
+  - list
+  - watch
```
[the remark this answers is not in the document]
zszabo-rh: Matches the existing pattern — all OSAC controllers share a single ClusterRole. Scoping Secrets RBAC to a namespaced Role would require splitting the service account, which is a broader change affecting all controllers. Tracked as a future hardening item, not blocking for this PR.

internal/controller/storage_controller.go
```diff
@@ -0,0 +1,740 @@
...
+
+func needsProvisionJob(latestJob *v1alpha1.JobStatus) bool {
+	if latestJob == nil || latestJob.JobID == "" {
+		return true
+	}
+	if latestJob.State == v1alpha1.JobStateFailed {
+		return false
+	}
+	return latestJob.State.IsSuccessful()
+}
```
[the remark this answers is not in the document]
zszabo-rh: Known issue, documented in E2E test findings as "stale job status." The controller may re-trigger after success if the expected resource hasn't appeared yet, but the AAP playbooks are idempotent so the redundant job is harmless. A timestamp-based backoff is a nice-to-have for post-v0.1.

internal/controller/storage_controller.go
```diff
@@ -0,0 +1,740 @@
...
+		requests = append(requests, reconcile.Request{
+			NamespacedName: client.ObjectKeyFromObject(&tenantList.Items[i]),
+		})
+	}
+	return requests
+}
+
+func (r *StorageReconciler) mapClusterOrderToTenant(_ context.Context, _ client.Object) []reconcile.Request {
+	return nil
+}
```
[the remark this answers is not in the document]
zszabo-rh: Intentional placeholder per the [design spec (PR #58)](https://github.com/osac-project/enhancement-proposals/pull/58). The ClusterOrder-to-Tenant mapping depends on the CaaS PRD (OSAC-1123). The no-op handler registers the watch so it's ready when the mapping is defined.

internal/controller/tenant_controller_test.go
```diff
@@ -18,817 +18,136 @@ package controller
...
-				reconcileUntilDeleting(ctx, nn)
-			}
-			ns := &corev1.Namespace{}
-			if err := k8sClient.Get(ctx, types.NamespacedName{Name: resourceName}, ns); err == nil {
-				_ = k8sClient.Delete(ctx, ns)
+			tenant := &v1alpha1.Tenant{}
+			if err := k8sClient.Get(ctx, typeNamespacedName, tenant); err == nil {
+				Expect(k8sClient.Delete(ctx, tenant)).To(Succeed())
 			}
 		})
```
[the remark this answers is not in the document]
zszabo-rh: Fixed in the latest commit — `BeforeEach` now handles `AlreadyExists` for the namespace.

internal/controller/tenant_controller_test.go
```diff
@@ -18,817 +18,136 @@ package controller
...
-			By("checking that a deprovision job was recorded")
-			Expect(k8sClient.Get(ctx, nn, tenant)).To(Succeed())
-			deprovJob := provisioning.FindLatestJobByType(tenant.Status.Jobs, v1alpha1.JobTypeDeprovision)
-			Expect(deprovJob).NotTo(BeNil())
-			Expect(deprovJob.JobID).To(Equal("mock-deprovision-job"))
+			tenant := &v1alpha1.Tenant{}
+			Expect(k8sClient.Get(ctx, typeNamespacedName, tenant)).To(Succeed())
+			Expect(tenant.Status.StorageClasses).To(BeNil())
+			Expect(tenant.Status.Jobs).To(BeNil())
 		})
```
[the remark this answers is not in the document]
zszabo-rh: Fair point. The test already proves the tenant controller doesn't touch storage by checking `StorageClasses` and `Jobs` are nil. Adding `StorageBackends` and `ClusterStorage` assertions would be more complete but the coverage is equivalent — the tenant controller has no code paths that set any of these fields.

internal/controller/storage_controller.go
```diff
@@ -0,0 +1,740 @@
...
+		return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
+	}
+
+	log.Info("backend provisioning job failed", "jobID", latestJob.JobID, "message", updatedJob.Message)
+	return ctrl.Result{}, nil
+}
+
+// --- Stage 2: Class provisioning ---
+
+func (r *StorageReconciler) handleClassProvisioning(ctx context.Context, instance *v1alpha1.Tenant) (ctrl.Result, error) {
```
akshaynadkarni: Nit: `handleClassProvisioning` / `handleClassDeprovisioning` / `pollClassProvisionJob` use "Class" terminology, but everything else in this PR uses "ClusterStorage" (condition `ClusterStorageReady`, job type `JobTypeClusterStorageProvision`, provider field `ClusterStorageProvider`, env vars `OSAC_STORAGE_CLUSTER_AAP_*`).

Looks like a leftover from before the design alignment commit renamed `StorageClassReady` to `ClusterStorageReady`. Consider renaming to `handleClusterStorageProvisioning` / `handleClusterStorageDeprovisioning` / `pollClusterStorageProvisionJob` for consistency.

internal/controller/storage_controller.go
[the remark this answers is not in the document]
akshaynadkarni: > A timestamp-based backoff is a nice-to-have for post-v0.1.

@zszabo-rh is there a ticket for this?

config/rbac/role.yaml
akshaynadkarni: Can we scope this with a namespaced Role + RoleBinding in `osac-system` instead of adding Secrets to the cluster-wide ClusterRole? That way the operator can only read Secrets in the namespace it actually needs, without changing the pod or service account.

Something like a `Role` in `osac-system` granting `secrets` get/list/watch, plus a `RoleBinding` pointing to the existing `controller-manager` ServiceAccount. The rest of the ClusterRole stays as-is.

WDYT?

internal/controller/storage_controller.go
```diff
@@ -0,0 +1,740 @@
...
+		if updateErr := r.Status().Update(ctx, instance); updateErr != nil {
+			return ctrl.Result{}, updateErr
+		}
+	}
+
+	log.Info("end storage reconcile")
+	return res, err
+}
+
+func (r *StorageReconciler) handleUpdate(ctx context.Context, instance *v1alpha1.Tenant) (ctrl.Result, error) {
```
akshaynadkarni, line 151: `StorageBackendStatus` and `ClusterStorageStatus` are defined on `TenantStatus` but never populated. Can we fill them in for VMaaS? For `ClusterStorageStatus`, the data is already available after `getTenantStorageClasses()` (target cluster name, ready state, reason). For `StorageBackendStatus`, a backend API is coming, but in the meantime the name/provider could be derived from the VAST config or env vars. WDYT?

internal/controller/storage_controller.go
akshaynadkarni, line 474: Nit: should we add a `// TODO(OSAC-1123)` here?

internal/controller/storage_controller.go
akshaynadkarni: The local `needsProvisionJob` re-triggers an AAP job when the previous one succeeded. But the failure case is already handled before this function is called ([lines 287-291](https://github.com/osac-project/osac-operator/blob/5b19dcafa97d6352ba07b08a3324ba74a1be3de0/internal/controller/storage_controller.go#L287-L291)), so the default `provisioning.NeedsProvisionJob` from `job_helpers.go` would work here:

- **No job**: triggers first job (same behavior)
- **Running**: falls through to poll (same behavior)
- **Failed**: never reaches this function, handled at [line 287](https://github.com/osac-project/osac-operator/blob/5b19dcafa97d6352ba07b08a3324ba74a1be3de0/internal/controller/storage_controller.go#L287) (same behavior)
- **Succeeded**: default returns false, falls through to `pollBackendProvisionJob`, which requeues after 5s. On the next reconcile, `hubSecretExists()` checks if the Secret appeared. No redundant AAP jobs.

Is there a reason the local function was needed instead of the shared one? WDYT?

internal/controller/tenant_names.go
```diff
@@ -53,3 +54,34 @@ var (
...
+				Operator: metav1.LabelSelectorOpIn,
+				Values:   []string{project},
+			},
+		},
+	}
+}
+
+func storageClassTenantPredicate() predicate.Predicate {
+	return predicate.NewPredicateFuncs(func(obj client.Object) bool {
+		_, exists := obj.GetLabels()[osacTenantAnnotation]
```
akshaynadkarni: Nit: `osacTenantAnnotation` is used as a label key everywhere (`client.MatchingLabels`, `obj.GetLabels()`), not as an annotation. Should this be renamed to `osacTenantLabel`?

internal/controller/storage_controller.go
zszabo-rh: You are absolutely right about this, fixing!

config/rbac/role.yaml
zszabo-rh: Good call, adding a namespaced Role + RoleBinding in osac-system for Secrets access and removing it from the ClusterRole.

internal/controller/storage_controller.go
zszabo-rh: I can populate ClusterStorageStatus from the existing data (cluster name from env, ready state from condition). StorageBackendStatus would need hardcoded values since there's no StorageBackend API yet, so feels a little premature. Happy to add ClusterStorageStatus population now and defer StorageBackendStatus to when the API lands, if that's ok with you.

internal/controller/storage_controller.go
zszabo-rh: Sure thing, adding TODO!

cmd/main.go
```diff
@@ -349,43 +357,79 @@ func setupComputeInstanceControllers(
...
 
 	aapURL := os.Getenv(envAAPURL)
 	aapToken := os.Getenv(envAAPToken)
 	if aapURL != "" && aapToken != "" {
-		tenantProvisionTemplate := helpers.GetEnvWithDefault(envTenantAAPProvisionTemplate, "osac-create-org")
-		tenantDeprovisionTemplate := helpers.GetEnvWithDefault(envTenantAAPDeprovisionTemplate, "osac-delete-org")
 		aapInsecureSkipVerify := helpers.GetEnvWithDefault(envAAPInsecureSkipVerify, false)
 
+		backendProvisionTemplate := helpers.GetEnvWithDefault(
+			envStorageBackendProvisionTemplate, "osac-create-tenant-storage-backend")
```
rgolangh, line 402: what does the default template do? can there be a default which will bring us to a functional state for storage?

internal/controller/storage_controller.go
zszabo-rh: Yes, seems I've badly overlooked this one.. Switching to `provisioning.NeedsProvisionJob`!

internal/controller/tenant_names.go
zszabo-rh: This one is coming from upstream, but I agree, renaming to `osacTenantLabel`.

internal/controller/storage_controller.go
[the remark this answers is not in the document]
zszabo-rh: Good point, but before creating new tickets, we should align the OSAC-23 epic structure with the merged design. The `/design:sync` phase was never run, so the current tickets (OSAC-1143 through OSAC-1146) were created before the design was finalized and do not reflect the final task breakdown. 
I'd suggest running the decompose + sync phases to get the epics/stories in shape first, then this backoff item would land in the right place.

cmd/main.go
zszabo-rh: No, the default doesn't bring you to a functional state on its own, it's just a convention that avoids extra env vars when using the standard AAP deployment (same pattern as all other controllers).

The full prerequisite chain is: controller enabled → AAP configured → templates deployed via config-as-code → STORAGE_TIERS env var set on instance group → VAST appliance reachable with valid credentials.

api/v1alpha1/job_types.go
```diff
@@ -21,14 +21,20 @@ import (
...
 	// JobTypeProvision indicates a provisioning operation
 	JobTypeProvision JobType = "provision"
 	// JobTypeDeprovision indicates a deprovisioning operation
 	JobTypeDeprovision JobType = "deprovision"
+
+	// Storage-specific job types used by the OSAC Storage Controller.
+	JobTypeStorageBackendProvision      JobType = "storage-backend-provision"
+	JobTypeStorageBackendDeprovision    JobType = "storage-backend-deprovision"
+	JobTypeClusterStorageProvision      JobType = "cluster-storage-provision"
+	JobTypeClusterStorageDeprovision    JobType = "cluster-storage-deprovision"
```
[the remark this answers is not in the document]
zszabo-rh: Known issue — this is the same concern Akshay raised. We're discussing two architectural alternatives (per-lifecycle CRDs or separate Jobs arrays) to resolve this. See the thread on the JobType issue comment.

config/crd/bases/osac.openshift.io_securitygroups.yaml
```diff
@@ -272,6 +272,10 @@ spec:
                       enum:
                       - provision
                       - deprovision
+                      - storage-backend-provision
+                      - storage-backend-deprovision
+                      - cluster-storage-provision
+                      - cluster-storage-deprovision
```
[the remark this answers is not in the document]
zszabo-rh: Same as above — the enum pollution is the symptom, the fix is architectural. Under discussion.

config/rbac/storage_secrets_role_binding.yaml
```diff
@@ -0,0 +1,16 @@
...
+  name: storage-secrets-reader-binding
+  namespace: osac-system
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: storage-secrets-reader
+subjects:
+- kind: ServiceAccount
+  name: controller-manager
+  namespace: system
```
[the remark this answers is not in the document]
zszabo-rh: False positive — `namespace: system` is the standard kustomize convention used by all existing RoleBindings in this project (leader_election, metrics_auth, admin, role_binding). The kustomize namespace transformer rewrites it to the actual deployment namespace.

internal/controller/storage_controller_test.go
```diff
@@ -0,0 +1,443 @@
...
+			Name: name,
+			Labels: map[string]string{
+				osacTenantKey:        tenant,
+				osacStorageTierLabel: tier,
+			},
+		},
+		Provisioner: "kubernetes.io/no-provisioner",
+	}
+	Expect(k8sClient.Create(ctx, sc)).To(Succeed())
+}
```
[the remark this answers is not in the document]
zszabo-rh: Fixed — added `DeferCleanup` to `createLabeledStorageClass` to delete the cluster-scoped StorageClass after each test.

internal/controller/storage_controller_test.go
```diff
@@ -0,0 +1,443 @@
...
+			Expect(k8sClient.Get(ctx, nn, tenant)).To(Succeed())
+			Expect(tenant.Finalizers).To(ContainElement(storageFinalizer))
+
+			Expect(k8sClient.Delete(ctx, tenant)).To(Succeed())
+
+			Eventually(func(g Gomega) {
+				_, err := r.Reconcile(ctx, storageReconcileRequest(nn))
+				g.Expect(err).NotTo(HaveOccurred())
+			}).Should(Succeed())
+		})
```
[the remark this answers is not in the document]
zszabo-rh: Fixed — the deletion test now asserts that the storage finalizer is removed from the Tenant after reconciliation.

internal/controller/storage_controller.go
```diff
@@ -0,0 +1,731 @@
...
+		}
+		return ctrl.Result{}, nil
+	}
+
+	instance.SetStatusCondition(v1alpha1.TenantConditionClusterStorageReady,
+		metav1.ConditionTrue,
+		v1alpha1.TenantReasonFound,
+		result.conditionMessage())
+	instance.Status.StorageClasses = result.resolved
+
```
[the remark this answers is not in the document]
zszabo-rh: Good catch — fixed. Added `instance.Status.StorageClasses = nil` on the unresolved path to clear stale entries when `ClusterStorageReady=False`.

internal/controller/storage_controller.go
```diff
@@ -0,0 +1,731 @@
...
+			return ctrl.Result{}, nil
+		case provisioning.DeprovisionWaiting:
+			return ctrl.Result{RequeueAfter: r.StatusPollInterval}, nil
+		default:
+			return ctrl.Result{RequeueAfter: r.StatusPollInterval}, nil
+		}
+	}
+
+	return r.pollDeprovisionJob(ctx, instance, latestJob, r.BackendProvider)
+}
```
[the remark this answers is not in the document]
zszabo-rh: The logic is safe in practice: `handleBackendDeprovisioning` checks `hubSecretExists()` first (line 506). It only returns an empty result (allowing finalizer removal) when the Secret is already gone AND no deprov job is running or blocking. The AAP teardown playbook only deletes the Secret after all VAST cleanup succeeds. But I agree an explicit re-check after job success would be a good defensive measure — will add in a follow-up.

internal/controller/tenant_controller.go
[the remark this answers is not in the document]
zszabo-rh: Fixed — updated to "Tracks namespace readiness and tenant lifecycle."

internal/controller/storage_controller_test.go
```diff
@@ -0,0 +1,451 @@
...
+				osacTenantKey:        tenant,
+				osacStorageTierLabel: tier,
+			},
+		},
+		Provisioner: "kubernetes.io/no-provisioner",
+	}
+	Expect(k8sClient.Create(ctx, sc)).To(Succeed())
+	DeferCleanup(func() {
+		_ = k8sClient.Delete(ctx, sc)
+	})
```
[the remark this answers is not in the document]
zszabo-rh: Fixed — using `client.IgnoreNotFound` for both the DeferCleanup delete and the deletion test Get, so only NotFound is accepted and other errors fail the test.

api/v1alpha1/clusterorder_types.go
```diff
@@ -150,11 +150,15 @@ type ClusterOrderStatus struct {
 	// NodeRequests reflects how many nodes are currently associated with the ClusterOrder
 	NodeRequests []NodeRequest `json:"nodeRequests,omitempty"`
 
-	// Jobs tracks the history of provision and deprovision operations
+	// ProvisioningJobs tracks the history of provision and deprovision operations
 	// Ordered chronologically, with latest operations at the end
 	// Limited to the last N jobs (configurable via OSAC_MAX_JOB_HISTORY, default 10)
 	// +kubebuilder:validation:Optional
-	Jobs []JobStatus `json:"jobs,omitempty"`
+	ProvisioningJobs []JobStatus `json:"provisioningJobs,omitempty"`
```
rgolangh, line 157: How long do we expect that array of jobs can get?
because we don't want to exceed the etcd object size limit so I suggest limit the array size.

zszabo-rh: as we already clarified in slack, this is controlled by an env var, and default job history size is 10
+etcd impact analysis from Claude (still using default, with ~200-300 bytes per entry): 

Realistic scenario — 100 tenants, each with 10 VMs, 5 subnets, 2 SGs, 1 cluster:
- 100 Tenants: 100 × 9 KB = 900 KB
- 100 ClusterOrders: 100 × 6 KB = 600 KB
- 1000 ComputeInstances: 1000 × 3 KB = 3 MB
- 500 Subnets: 500 × 3 KB = 1.5 MB
- 100 VirtualNetworks: 100 × 3 KB = 300 KB
- 200 SecurityGroups: 200 × 3 KB = 600 KB

Total job history across all CRs: ~7 MB in etcd — spread across thousands of individual objects (each well under the 1.5MB per-object limit).

Risk assessment: No etcd concern. The job history cap of 10 per array keeps each individual CR small (~3-9 KB of job data). Even at scale, the total is modest compared to typical etcd usage. The only scenario that could approach the per-object limit is if message fields contain unusually large error strings — but AAP messages are typically short.

rgolangh: good, thanks for checking.

internal/controller/storage_controller.go
```diff
@@ -0,0 +1,496 @@
...
+)
+
+const (
+	storageFinalizer      = "osac.openshift.io/storage"
+	storageControllerName = "storage-controller"
+)
+
+// StorageReconciler reconciles storage lifecycle on Tenant CRs.
+// It owns StorageBackendReady, ClusterStorageReady conditions,
+// status.storageClasses, and status.jobs on the Tenant CR.
```
akshaynadkarni: Nit: this still references `status.jobs`. Should be `status.storageBackendJobs` and `status.clusterStorageJobs`.

pkg/provisioning/provision_lifecycle.go
```diff
@@ -96,13 +74,15 @@ func EvaluateAction(provState *State, checkAPIServer func() bool) (Action, *v1al
 }
 
 // CheckAPIServerForNonTerminalProvisionJob reads the resource directly from the API server
-// and returns true if a non-terminal provision job exists.
-func CheckAPIServerForNonTerminalProvisionJob(ctx context.Context, apiReader client.Reader, key client.ObjectKey, fresh client.Object) bool {
+// and returns true if a non-terminal provision job exists. The extract parameter determines
+// which jobs array to check — standard controllers pass GetJobsFromResource, while the
```
akshaynadkarni: Nit: this still references `GetJobsFromResource`. Should be `JobsExtractor`.

internal/controller/storage_controller.go
zszabo-rh: done!

pkg/provisioning/provision_lifecycle.go
zszabo-rh: done and fixed a few other stale references as well, thanks for finding these!

config/rbac/storage_secrets_role.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  labels:
+    app.kubernetes.io/name: osac-operator
+    app.kubernetes.io/managed-by: kustomize
+  name: storage-secrets-reader
+  namespace: osac-system
```
akshaynadkarni: The Role and RoleBinding hardcode `namespace: osac-system`, but the Helm chart deploys to `{{ .Release.Namespace }}` (typically `osac`). The [RoleBinding](https://github.com/osac-project/osac-operator/blob/566e8d4b733493856121d8ed8f9279ecddaf7268/config/rbac/storage_secrets_role_binding.yaml#L16) also references the SA in `namespace: system`, which Helm doesn't transform.

All other namespace env vars in the [Helm deployment template](https://github.com/osac-project/osac-operator/blob/566e8d4b733493856121d8ed8f9279ecddaf7268/charts/operator/templates/deployment.yaml#L65-L84) use `metadata.namespace` via the downward API. Storage should follow the same pattern:

1. **`config/rbac/storage_secrets_role.yaml`**: Change `namespace: osac-system` to `namespace: system` so kustomize transforms it like the other RBAC resources
2. **[`config/rbac/storage_secrets_role_binding.yaml#L8`](https://github.com/osac-project/osac-operator/blob/566e8d4b733493856121d8ed8f9279ecddaf7268/config/rbac/storage_secrets_role_binding.yaml#L8)**: Same, change `namespace: osac-system` to `namespace: system`
3. **[`charts/operator/values.yaml`](https://github.com/osac-project/osac-operator/blob/566e8d4b733493856121d8ed8f9279ecddaf7268/charts/operator/values.yaml)**: Add a storage config namespace value (default: release namespace)
4. **[`charts/operator/templates/deployment.yaml`](https://github.com/osac-project/osac-operator/blob/566e8d4b733493856121d8ed8f9279ecddaf7268/charts/operator/templates/deployment.yaml#L65-L84)**: Add `OSAC_STORAGE_CONFIG_NAMESPACE` using `metadata.namespace`, same as `OSAC_TENANT_NAMESPACE` and the others
5. **`charts/operator/templates/`**: Add Role + RoleBinding templates using `{{ .Release.Namespace }}`
6. **[`storage_controller.go#L392-L396`](https://github.com/osac-project/osac-operator/blob/566e8d4b733493856121d8ed8f9279ecddaf7268/internal/controller/storage_controller.go#L392-L396)**: The `osac-system` fallback in `storageConfigNamespace()` is fine for local dev, but a comment would help clarify that Helm/kustomize should always set the env var:

```go
// storageConfigNamespace returns the namespace where storage config Secrets
// are stored. In production, OSAC_STORAGE_CONFIG_NAMESPACE is set by the
// Helm chart via the downward API to match the deployment namespace.
// The osac-system fallback is for local development only.
```

WDYT?

internal/controller/storage_controller_test.go
```diff
@@ -0,0 +1,581 @@
...
+				g.Expect(err).NotTo(HaveOccurred())
+
+				g.Expect(k8sClient.Get(ctx, nn, tenant)).To(Succeed())
+				cond := tenant.GetStatusCondition(v1alpha1.TenantConditionStorageBackendReady)
+				g.Expect(cond).NotTo(BeNil())
+				g.Expect(cond.Status).To(Equal(metav1.ConditionFalse))
+			}, 5*time.Second, 100*time.Millisecond).Should(Succeed())
+		})
+	})
+})
```
akshaynadkarni, line 671: Can we add unit tests for:
1. Deprovisioning ordering (cluster storage cleaned up before backend teardown)
2. Duplicate StorageClass detection (multiple SCs for the same tier should set ClusterStorageReady=False with MultipleFound reason)

The test helpers and mock provider are already in place, so these should be straightforward to add.

internal/controller/storage_controller.go
akshaynadkarni: Makes sense to populate `ClusterStorageStatus` now and defer `StorageBackendStatus`. Can we add a TODO referencing OSAC-1111 where `StorageBackendStatus` is defined, so it's clear this is deferred until the backend API exists?

config/rbac/storage_secrets_role.yaml
zszabo-rh: wow, good catch!! 
fixed it, changed kustomize Role/RoleBinding to `namespace: system` for proper transformation, added Helm templates with `{{ .Release.Namespace }}`, added `OSAC_STORAGE_CONFIG_NAMESPACE` env var to Helm deployment (same downward API pattern as other namespace vars), and added a comment to `storageConfigNamespace()` clarifying the fallback.

internal/controller/storage_controller_test.go
zszabo-rh: done!

internal/controller/storage_controller.go
zszabo-rh: sure thing, done!

